PR #20, opened 2019-12-08
Add Remote Player File Installation

drinfernoo:
Using this, it is possible to input a URL to a `.zip` file containing player files, and have them extracted into the `addon_data` player file location.

A convenience of this is that it works very with [GitHub's zipball endpoint](https://developer.github.com/v3/repos/contents/#get-archive-link), which allows a URL like: `https://www.github.com/drinfernoo/json.tmdbplayers/zipball/master/` to work perfectly. The aforementioned link is to my test repo, which currently only includes the same player files that are included with this add-on, but should function as a proof-of-concept.

The `.zip` file will be extracted "flat", ignoring any folder structure contained inside, and only extract files ending in `.json`.

### DIFF
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -267,7 +267,7 @@ def open_url(url, stream=False, check=False, cred=None, count=0):
             return True if valid else False
             
         if valid == 'auth' and not cred:
-            cred = (get_keyboard(heading='Username'), get_keyboard(heading='Password'))
+            cred = (xbmcgui.Dialog().input(heading='Username') or '', xbmcgui.Dialog().input(heading='Password', option=xbmcgui.ALPHANUM_HIDE_INPUT) or '')
             
         response = requests.get(url, timeout=10.000, stream=stream, auth=cred)
 
@@ -276,7 +276,7 @@ def open_url(url, stream=False, check=False, cred=None, count=0):
             
             if retry and count < 3:
                 count += 1
-                cred = (get_keyboard(heading='Username'), get_keyboard(heading='Password'))
+                cred = (xbmcgui.Dialog().input(heading='Username') or '', xbmcgui.Dialog().input(heading='Password', option=xbmcgui.ALPHANUM_HIDE_INPUT) or '')
                 
                 response = open_url(url, stream, check, cred, count)
             else:
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -200,3 +200,87 @@ def make_kwparams(params):
     return del_dict_keys(tempparams, ['info', 'type', 'tmdb_id', 'filter_key', 'filter_value',
                                       'with_separator', 'with_id', 'season', 'episode', 'prop_id',
                                       'exclude_key', 'exclude_value'])
+                                      
+
+def get_keyboard(default="", heading="", hidden=False):
+    keyboard = xbmc.Keyboard(default, heading, hidden)
+    keyboard.doModal()
+    if keyboard.isConfirmed():
+        return keyboard.getText()
+    return default
+                                      
+                                      
+def _is_url(url):
+    try:  # Python 3
+        from urllib.parse import urlparse
+    except ImportError:  # Python 2
+        from urlparse import urlparse
+
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
+        
+
+def _check_url(url, cred):
+    import requests
+
+    if _is_url(url):
+        try:
+            response = requests.head(url, allow_redirects=True, auth=cred)
+            
+            if response.status_code < 300:
+                xbmc.log("URL check passed for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
+                return True
+            elif response.status_code < 400:
+                xbmc.log("URL check redirected from {0} to {1}: Status code [{2}]".format(url, response.headers['Location'], response.status_code), level=xbmc.LOGDEBUG)
+                return _check_url(response.headers['Location'])
+            elif response.status_code == 401:
+                xbmc.log("URL requires authentication for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
+                return 'auth'
+            else:
+                xbmc.log("URL check failed for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
+                return False
+        except Exception as e:
+            xbmc.log("URL check error for {0}: [{1}]".format(url, e), level=xbmc.LOGDEBUG)
+            return False
+    else:
+        xbmc.log("URL is not of a valid schema: {0}".format(url), level=xbmc.LOGDEBUG)
+        return False
+        
+
+def open_url(url, stream=False, check=False, cred=None, count=0):
+    import requests
+
+    if not url:
+        return False
+
+    count = 0
+    
+    valid = _check_url(url, cred)
+
+    if not valid:
+        return False
+    else:
+        if check:
+            return True if valid else False
+            
+        if valid == 'auth' and not cred:
+            cred = (get_keyboard(heading='Username'), get_keyboard(heading='Password'))
+            
+        response = requests.get(url, timeout=10.000, stream=stream, auth=cred)
+
+        if response.status_code == 401:
+            retry = xbmcgui.Dialog().yesno(_addon.getAddonInfo('name'), 'Either the username or password were invalid. Would you like to try again?', yeslabel='Retry', nolabel='Cancel')
+            
+            if retry and count < 3:
+                count += 1
+                cred = (get_keyboard(heading='Username'), get_keyboard(heading='Password'))
+                
+                response = open_url(url, stream, check, cred, count)
+            else:
+                xbmcgui.Dialog().ok(_addon.getAddonInfo('name'), 'Authentication Failed.')
+                return False
+        
+        return response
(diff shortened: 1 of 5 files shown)


jurialmunkey:
Looks interesting!  :grin:

However, I noticed a few issues:
There's an error in script.py
Also, even with out the error, the function is never called - only the log is written to. 
Plus you need to return after the log write because otherwise script.py will call a window.

Can't use shutil module in official repo - need to use xbmcvfs. Same for os module afaik.

Don't do `except: pass`
At the very least, you need to pass the Exception to the log - but, even then, its bad practice. Only place I do that is for Service Monitor because there isn't a way to restart it other than restarting Kodi.

Not keen on modules being imported inside functions. Just import the necessary module at the top of the file.

I have a logging function in `utils.kodi_log(string)` - use that for logging so that the log is correctly prefixed with the addon name. Do `utils.kodi_log(string, 0)` for debug logging and `utils.kodi_log(string, 1)` for explicit logging.

drinfernoo:
Oh shoot! I didn't mean to leave that log statement, must've missed a commit 😅

I can make the changes to imports, exceptions, and logging, though 😁

jurialmunkey:
>
>
> Does the official repo not allow the use of `os` or `shutil`? That seems like an arbitrary limitation, since they're part of the standard Python libraries 🤔

Actually, I think `os` is fine. However, `shutil` isn't recommended - it can be used if absolutely necessary but where possible you should use xbmcvfs instead to ensure cross-platform compatibility. Plus it has better security because it limits access to the file system, which is important considering files are going to be downloaded from an online source.

I can't see any functions that you are using in shutil that can't be done with xbmcvfs:
https://codedocs.xyz/AlwinEsch/kodi/group__python__xbmcvfs.html

Also I have to double check if downloading files in this manner is allowed (I'm pretty sure it should be fine though). The alternative would be to add a url as a source in kodi file manager and then use built-in file browser dialog.

drinfernoo:
So I didn't need to fix logging, because that old log statement wasn't even needed. On the topic of `shutil` vs `xbmcvfs`, though... I need to take another pass, as `shutil` takes `File` objects, but `xbmcvfs` takes `str`s.

jurialmunkey:
No worries. Ping me when you're done and I'll do a review.

BTW, that get_keyboard function can be replaced with a call straight to `xbmcgui.Dialog().input()`
https://codedocs.xyz/AlwinEsch/kodi/group__python___dialog.html#ga645b30a9a82e8f113e34859db85db658

Real example replacing get_keyboard() function:
```
cred = (xbmcgui.Dialog().input(heading='Username') or '', xbmcgui.Dialog().input(heading='Password', option=xbmcgui.ALPHANUM_HIDE_INPUT) or '')
```

drinfernoo:
@jurialmunkey I think I've made all the requested changes :+1:

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+        success = recursive_delete_dir(os.path.join(fullpath, directory))
+    success = xbmcvfs.rmdir(fullpath)
+    return success
+                                      
+                                      
+def _is_url(url):
+    try:  # Python 3
+        from urllib.parse import urlparse
+    except ImportError:  # Python 2
+        from urlparse import urlparse
```
jurialmunkey: Move imports to top of file.

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
+        
+
+def _check_url(url, cred):
+    import requests
```
jurialmunkey: Move imports to top of file.

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+
+    if _is_url(url):
+        try:
+            response = requests.head(url, allow_redirects=True, auth=cred)
+            
+            if response.status_code < 300:
+                xbmc.log("URL check passed for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
```
jurialmunkey: Use kodi_log() method.

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+                return True
+            elif response.status_code < 400:
+                xbmc.log("URL check redirected from {0} to {1}: Status code [{2}]".format(url, response.headers['Location'], response.status_code), level=xbmc.LOGDEBUG)
```
jurialmunkey: kodi_log()

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+                return _check_url(response.headers['Location'])
+            elif response.status_code == 401:
+                xbmc.log("URL requires authentication for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
```
jurialmunkey: kodi_log()

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+                return 'auth'
+            else:
+                xbmc.log("URL check failed for {0}: Status code [{1}]".format(url, response.status_code), level=xbmc.LOGDEBUG)
```
jurialmunkey: kodi_log()

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+                return False
+        except Exception as e:
+            xbmc.log("URL check error for {0}: [{1}]".format(url, e), level=xbmc.LOGDEBUG)
```
jurialmunkey: kodi_log()

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+            return False
+    else:
+        xbmc.log("URL is not of a valid schema: {0}".format(url), level=xbmc.LOGDEBUG)
```
jurialmunkey: kodi_log()

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+
+    count = 0
+    
+    valid = _check_url(url, cred)
+
+    if not valid:
+        return False
+    else:
+        if check:
+            return True if valid else False
```
jurialmunkey: You already returned False if not valid so this line will always return True.

You could also drop the the else statement and just do:
```
if not valid:
    return False
if check:
    return True
[[rest of code]]
```

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+        return False
+        
+
+def open_url(url, stream=False, check=False, cred=None, count=0):
+    import requests
+
+    if not url:
+        return False
+
+    count = 0
```
jurialmunkey: This line defeats the purpose of doing a retry count because you reset it to zero every time.

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+            cred = (xbmcgui.Dialog().input(heading='Username') or '', xbmcgui.Dialog().input(heading='Password', option=xbmcgui.ALPHANUM_HIDE_INPUT) or '')
+            
+        response = requests.get(url, timeout=10.000, stream=stream, auth=cred)
+
+        if response.status_code == 401:
+            retry = xbmcgui.Dialog().yesno(_addon.getAddonInfo('name'), 'Either the username or password were invalid. Would you like to try again?', yeslabel='Retry', nolabel='Cancel')
+            
+            if retry and count < 3:
+                count += 1
+                cred = (xbmcgui.Dialog().input(heading='Username') or '', xbmcgui.Dialog().input(heading='Password', option=xbmcgui.ALPHANUM_HIDE_INPUT) or '')
```
jurialmunkey: Your retry counter never hits 3 because you reset to zero earlier in the function.

resources/settings.xml
```diff
@@ -4,10 +4,13 @@
...
+        <setting label="$ADDON[plugin.video.themoviedb.helper 32010]" type="action" action="RunScript(plugin.video.themoviedb.helper, authenticate_trakt)" option="close" />
+        <setting label="$ADDON[plugin.video.themoviedb.helper 32013]" type="bool" id="trakt_management" default="false" />
+        <setting label="$ADDON[plugin.video.themoviedb.helper 32014]" type="lsep" />
+        <setting label="$ADDON[plugin.video.themoviedb.helper 32015]" type="text" id="players_url" default="" />
+        <setting label="$ADDON[plugin.video.themoviedb.helper 32016]" type="action" action="RunScript(plugin.video.themoviedb.helper, update_players)" subsetting="true" option="close" enabled="!eq(-1,)" />
     </category>
```
jurialmunkey: Settings are getting a bit crowded - might be worth sticking these download settings in their own category.

resources/lib/utils.py
```diff
@@ -200,3 +200,95 @@ def make_kwparams(params):
...
+        except Exception as e:
+            xbmc.log("URL check error for {0}: [{1}]".format(url, e), level=xbmc.LOGDEBUG)
+            return False
+    else:
+        xbmc.log("URL is not of a valid schema: {0}".format(url), level=xbmc.LOGDEBUG)
+        return False
+        
+
+def open_url(url, stream=False, check=False, cred=None, count=0):
+    import requests
```
jurialmunkey: Move import to top.

jurialmunkey:
@drinfernoo - Don't worry about my requested changes. I've made them locally and I'm about to merge this PR - Just a heads-up so we don't double up on the same work!

drinfernoo:
👍

jurialmunkey:
Merged.

I ended up sticking the code into its own module as I figure it might have other uses down the line.

PR #22, opened 2019-12-13
Default players

drinfernoo:
This allows for user-settable default players for each mediatype. Simply set them (from a list of applicable players) using the action in settings, and "playing" that content from the add-on will automatically attempt to use the specified player. If default players have not yet been defined, or have been cleared, the select dialog will still be shown to choose a player.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -422,3 +422,8 @@ msgstr ""
 msgctxt "#32026"
 msgid "Choose default players"
 msgstr ""
+
+#: /resources/settings.xml
+msgctxt "#32027"
+msgid "Clear default players"
+msgstr ""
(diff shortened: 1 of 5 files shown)


jurialmunkey:
Awesome! Thanks! 😁 

Works great. I made a few minor tweaks/improvements - mainly ability to set/clear defaults individually.

Also, FYI you can add text of a setting to an action command simply by adding the id:
`<setting label="$ADDON[plugin.video.themoviedb.helper 32024]" type="action" action="RunScript(plugin.video.themoviedb.helper, set_defaultplayer=movie)" id="default_player_movies" />`

drinfernoo:
@jurialmunkey Nice! That was the only reason I *didn't* have them be settable separately 🤦 Glad this one worked out 👍

jurialmunkey:
@drinfernoo - Yeah it's not in the docs on the kodi wiki. Thought I would just try see if it works and it did!

drinfernoo:
@jurialmunkey I dig it! Thanks for the tip!

PR #33, opened 2019-12-21
:gear: - Show players URL in settings

drinfernoo:


jurialmunkey:
I was debating about adding this myself - wasn't sure if there was a point because the space to show the url was pretty small and it would have to scroll. Still probably better to show something there rather than nothing and that way it will indicate that it has already been set. Merged. :D

PR #39, opened 2019-12-28
:children_crossing: - Hide context options on non-TMDBHelper content

drinfernoo:


jurialmunkey:
Does `String.IsEqual(Container.PluginName,plugin.video.themoviedb.helper)` work instead of the `String.StartsWith(ListItem.Path,plugin://plugin.video.themoviedb.helper)` condition?

String.StartsWith is more expensive than IsEqual. 

Plus, I'm planning on pointing the path directly at the local file if the item is in the local dd and that would break this condition for local db items.

jurialmunkey:
Nevermind, I just tested with `Container.PluginName` and it doesn't work with widgets, so I'll merge this one instead.

PR #49, opened 2020-01-19
Added new methods, play from tvdb_id/query by name

henryjfry:
Modified "def list_play(self):" to allow playing a movie or an episode from the name or the tvdb id which will allow triggering from external addons eg from a context menu item in openinfo.

So the following commands will play via tmdb helper:

kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;tvdb_id=70726&amp;type=episode&amp;season=5&amp;episode=1)'

kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;query=Mr.+Robot&amp;type=episode&amp;season=4&amp;episode=13)'

kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;query=The Abyss&amp;year=1989)'

### DIFF
--- a/resources/lib/container.py
+++ b/resources/lib/container.py
@@ -584,11 +584,33 @@ class Container(Plugin):
         self.params['tmdb_id'] = self.get_tmdb_id(**self.params)
 
     def list_play(self):
-        if not self.params.get('type') or not self.params.get('tmdb_id'):
+	    tmdb_id = ''
+	    if self.params.get('query') and self.params.get('type') == 'episode':
+		    tmdb_id = self.tmdb.get_tmdb_id(itemtype='tv', query=self.params.get('query'))
+
+	    if self.params.get('query') and self.params.get('type') == 'movie':
+		    movie_title = self.params.get('query')
+		    movie_year = self.params.get('year')
+		    tmdb_id = self.tmdb.get_tmdb_id(itemtype='movie', query=self.params.get('query'))
+
+	    if self.params.get('tvdb_id') and self.params.get('type') == 'episode':
+		    tvdb_id = self.params.get('tvdb_id')
+		    tmdb_id = self.tmdb.get_tmdb_id(itemtype='tv', tvdb_id=tvdb_id)
+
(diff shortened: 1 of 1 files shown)


jurialmunkey:
The script was meant to already do this but there was a bug because it was trying to convert ids/queries for episodes rather than base tvshow. Fixed in latest: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/285804a25a08caaab0d16cb44ccc8a9c006622d0

All that needed to be changed was the `list_getid()` function
From this:
```python
def list_getid(self):
    self.params['tmdb_id'] = self.get_tmdb_id(**self.params)
```

To this:
```python
def list_getid(self):
    params = self.params.copy()
    if self.params.get('info') == 'play' and self.params.get('type') == 'episode':
        params['type'] = 'tv'
    self.params['tmdb_id'] = self.get_tmdb_id(**params)
```

henryjfry:
So with that change will the queries work exactly as is?
eg:
kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;tvdb_id=70726&amp;type=episode&amp;season=5&amp;episode=1)'

kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;query=Mr.+Robot&amp;type=episode&amp;season=4&amp;episode=13)'

kodi-send --action='RunPlugin(plugin://plugin.video.themoviedb.helper?info=play&amp;query=The Abyss&amp;year=1989)'

So i know if i need to make amendments to my context menu items.

jurialmunkey:
Yes they should do. As long as the tvdb id being passed is for the tvshow and not the episode. tmdb has episode ids in addition to tvshow ids (not sure if tvdb is the same) - I use tvshow id for player so any query/id lookup needs to be for the tvshow name not  episode name

henryjfry:
Cool, thanks for confirming.

PR #57, opened 2020-01-25
Fix search for TV shows

jandurovec:
API for TV show search doesn't have 'year' parameter so the currently
passed parameter is ignored. As a result of that the helper returns
invalid information for TV shows where query by name is ambiguous.

The correct parameter name is first_air_date_year.


jurialmunkey:
Looks good. Thanks!

PR #78, opened 2020-02-07
Fix bug with reused TMDb IDs

drinfernoo:
This will fix the instance where the played item has the same TMDb ID as another item of a different mediatype, for example:

https://www.themoviedb.org/tv/251-aqua-teen-hunger-force
https://www.themoviedb.org/movie/251-ghost

This PR makes sure that the correct item will be chosen in the process of building details for the item, including slugs, IDs, etc...

### DIFF
--- a/resources/lib/traktapi.py
+++ b/resources/lib/traktapi.py
@@ -389,7 +389,8 @@ class TraktAPI(RequestAPI):
 
     def get_traktslug(self, item_type, id_type, id_num):
         item = self.get_response_json('search', id_type, id_num, '?' + item_type)
-        return item[0].get(item_type, {}).get('ids', {}).get('slug')
+        for entry in [x for x in item if item_type in x]:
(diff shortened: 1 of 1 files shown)


SerpentDrago:
I can confirm this PR Fix's the issue i was having with some addons players not working for some shows. (the addon in question uses trakt id's in this case) 

thanks Dr !

jurialmunkey:
@drinfernoo - looks good. Have you tested what happens if None type is returned (ie no match)?

jurialmunkey:
Okay I think it should be fine if in the random chance None is returned.

drinfernoo:
> @drinfernoo - looks good. Have you tested what happens if None type is returned (ie no match)?

As long as `item is not None`, then there should be no `Exception`, even if there is no match. Even in that case before, there was no `Exception` from TMDb Helper, but could be potentially thrown from the playing add-on.

PR #115, opened 2020-03-01
add no_folder player option

L2501:
no_folder player option for players that return one playable item via setResolvedUrl

### DIFF
--- a/resources/lib/player.py
+++ b/resources/lib/player.py
@@ -71,6 +71,10 @@ class Player(Plugin):
             actionlist = player[1]
             player = (False, actionlist[0])
             for d in actionlist[1:]:
+                if d.get('no_folder'):
+                    resolve_url = True
+                    player = (resolve_url, actionlist[0])
+
                 if player[0]:
                     break  # Playable item was found in last action so let's break and play it
                 folder = KodiLibrary().get_directory(string_format_map(player[1], self.item))  # Get the next folder from the plugin


jurialmunkey:
What's the use case for this action? Why is this step necessary? If the path only returns one item and no further actions are defined, then the player will play the file.

The only time I could see this being useful is if the item return has the isFolder flag set to True. And in that case, it shouldn't be playable because it is a folder...

jurialmunkey:
For instance, Youtube uses setResolveUrl and I can play the first item returned without a problem using this:
```
"play_movie"        : [
                        "plugin://plugin.video.youtube/search/?q={name}&search_type=notvalid",
                        {"position": "1"}
                      ],
```

L2501:
/search/ returns a folder not a playable item

a playable item is `plugin://plugin.video.youtube/play/?video_id=dQw4w9WgXcQ` for example

jurialmunkey:
> 
> 
> /search/ returns a folder not a playable item
> 
> a playable item is `plugin://plugin.video.youtube/play/?video_id=dQw4w9WgXcQ` for example

You can already play a path like this using
`"play_movie" : "plugin://plugin.video.youtube/play/?video_id=dQw4w9WgXcQ"`

I tested a youtube path like that already and it works completely fine on my end. The xbmc player can take either a plugin url that needs to be resolved or the actual resolved filepath - both are accepted.

jurialmunkey:
BTW you realise that there is a difference between using `search_movie` and `play_movie`? 
Play movie expects a resolvable url where as search expects a folder.

L2501:
i see, wiki should be then

|Key|Description|
|---|---|
|play_movie | movie plugin url passed to `RunPlugin(url)` if list or `PlayMedia(url)` if string|
|play_play_episode | episode plugin url passed to `RunPlugin(url)` if list or `PlayMedia(url)` if string|

PR #177, opened 2020-04-20
Add Embycon player

juanma-cvega:
The author of the Embycon addon just merged my changes to his branch so now the new search player that uses the TV show TVDB ID should work. 
I added an assert as well to avoid using the play_episode player when there is no TVDB ID and a fallback to try using the new method when the match by title and premiered date doesn't work. 
I added the fallback for movies as well.


jurialmunkey:
Awesome, thanks! This is great!

PR #216, opened 2020-05-12
Add spanish translation [es_ES]

roliverosc:
Hi @jurialmunkey I have allowed myself translate this addon to spanish. I would like it to be available for other users.


jurialmunkey:
Awesome! Thanks! 😁

PR #217, opened 2020-05-12
add player for Composite

anxdpanic:
This requires a beta version of Composite currently. 
https://github.com/anxdpanic/plugin.video.composite_for_plex/releases/tag/1.1.0-dev

Is there a way to create a fallback player that doesn't show up in the dialog?

I'd prefer to remove the `{"dialog": "auto"}` from the search endpoints, and create a fallback player for that which doesn't show in that dialog.

Allowing custom named fallback keys could resolve it as well,
ie.
```
{
    "play_movie": "composite_for_plex.json play_movie_2",
    "play_episode": "composite_for_plex.json fb_play_episode"
}
```

### DIFF
--- a/resources/players/composite_for_plex.json
+++ b/resources/players/composite_for_plex.json
@@ -1,0 +1,31 @@
+{
+    "name"              : "Composite",
+    "plugin"            : "plugin.video.composite_for_plex",
+    "priority"          : 200,
+    "assert"            : {
+                            "play_movie":       ["title", "year"],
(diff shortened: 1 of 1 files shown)


jurialmunkey:
Awesome! Thanks 😁

jurialmunkey:
> Is there a way to create a fallback player that doesn't show up in the dialog?

Not currently, but I can see the value of some type of "fallback only" type setting that would hide it from the dialog. I'll keep you posted, shouldn't be too hard to implement.

PR #246, opened 2020-06-03
:sparkles: - Allow multiple plugins in player file

### LINKED ISSUE #245: [Request] Allow Multiple 'plugin' Values in Player
Simply put, it would be great if the `plugin` key in a player file was (or could be) a list, so that player files which declare fallbacks can be hidden when the add-ons aren't both installed.

For example, if I want a player that will play from Netflix, but fall back to Disney+ (hypothetically, that player file has yet to be seen), then that player file should *only* be displayed in the selector when *both* `plugin.video.netflix` and `slyguy.disney.plus` are installed.

drinfernoo:
Close #245

### DIFF
--- a/resources/lib/player.py
+++ b/resources/lib/player.py
@@ -305,8 +305,17 @@ class Player(Plugin):
                     vfs_file.close()
 
                 self.players[file] = meta
-                if not meta.get('plugin') or not xbmc.getCondVisibility(u'System.HasAddon({0})'.format(meta.get('plugin'))):
-                    continue  # Don't have plugin so skip
+                plugin = meta.get('plugin')
+                missing = False
+                if plugin:
+                    if isinstance(plugin, list):
+                        for p in plugin:
+                            if not xbmc.getCondVisibility(u'System.HasAddon({0})'.format(p)):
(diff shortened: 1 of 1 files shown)


jurialmunkey:
@drinfernoo - Awesome! Thanks.

You never actually checked `if not plugin` so it would build the player if the plugin field was empty. Really minor issue though, as that's really just included as a sanity check to make sure the file actually is a player file and is somewhat configured correctly.

FYI - you can use a `for ... else` loop to run a command only if the condition is true for every item in the list. Breaking a for loop means that the `else` section won't run - see e.g.  https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/4eb55529a594c4cdcac78666ef74f0a9db2fb214

drinfernoo:
Wow, yours is way cleaner! Thanks!

jurialmunkey:
> 
> 
> Wow, yours is way cleaner! Thanks!

I only thought to do it this way because I was messing around in the player stuff before and part of the code for the player steps uses this `for ... else` approach. Using else with a for loop feels weird and back-to-front so I always forget that it's possible.

PR #267, opened 2020-07-07
:alien::bug: - Fix Trakt Authorization Dialog

drinfernoo:
`xbmcgui.DialogProgress.create(...)` no longer takes three lines of text. It only takes a header and a single multiline text.

### DIFF
--- a/resources/lib/traktapi.py
+++ b/resources/lib/traktapi.py
@@ -59,7 +59,7 @@ class TraktAPI(RequestAPI):
         self.auth_dialog = xbmcgui.DialogProgress()
         self.auth_dialog.create(
             self.addon.getLocalizedString(32097),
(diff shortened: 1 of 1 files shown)


jurialmunkey:
Must've missed the notification for this PR.

Lgtm. I'm guessing the other modal dialogs (yesno and ok) probably require this change for Matrix as well.

FYI - PEP8 guidelines _now_ recommend breaking _before_ operators:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

I think a lot of my code followed the old way of breaking after operators - I'm trying to transition to the new guideline now though as imho I do think it is clearer. Not super important as it works both ways, just thought I'd point out the new guidelines as I wasn't aware of them for a long time.

PR #353, opened 2020-11-09
python2 fix for utf-8 encoding of paths passed to xbmc.executebuiltin()

dplim87:
Fixes issue with players which include utf-8 characters in path.


jurialmunkey:
Excellent, thanks! lgtm.

PR #372, opened 2020-11-20
proposed tweaks

dplim87:
Trying to call py3 functions directly and use shims for py2 and also some changes to avoid unicode conversion errors in py2

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon id="plugin.video.themoviedb.helper"
-version="4.0.59"
+version="4.0.60"
 name="TheMovieDb Helper"
 provider-name="jurialmunkey">
 <requires>
--- a/resources/lib/addon/timedate.py
+++ b/resources/lib/addon/timedate.py
@@ -1,3 +1,4 @@
+import _strptime
 import datetime
 import xbmc
 import time
--- a/resources/lib/trakt/progress.py
+++ b/resources/lib/trakt/progress.py
@@ -5,7 +5,7 @@ from resources.lib.trakt.items import TraktItems
 from resources.lib.trakt.decorators import is_authorized, use_activity_cache, use_lastupdated_cache
 from resources.lib.addon.parser import try_int
 from resources.lib.addon.timedate import convert_timestamp, date_in_range, get_region_date, get_datetime_today, get_timedelta
-from resources.lib.addon.plugin import viewitems, kodi_log
+from resources.lib.addon.plugin import viewitems
 from resources.lib.api.mapping import get_empty_item
 # from resources.lib.addon.decorators import timer_report
 
@@ -270,9 +270,6 @@ class _TraktProgress():
     def _get_calendar_episodes_list(self, startdate=0, days=1, user=True, kodi_db=None):
         # Get response
         response = self.get_calendar_episodes(startdate=startdate, days=days, user=user)
-        kodi_log([
-            'get_calendar_episodes', ' startdate=', startdate, ' days=', days, ' user=', user,
-            '\n', 'kodi_db=', 'True' if kodi_db else 'False', ' response=', len(response)], 1)
         if not response:
             return
         # Reverse items for date ranges in past
@@ -282,7 +279,6 @@ class _TraktProgress():
 
     def get_calendar_episodes_list(self, startdate=0, days=1, user=True, kodi_db=None, page=1, limit=20):
         response_items = self._get_calendar_episodes_list(startdate, days, user, kodi_db)
-        kodi_log(['get_calendar_episodes', ' response_items=', len(response_items)], 1)
         response = PaginatedItems(response_items, page=page, limit=limit)
         if response and response.items:
             return response.items + response.next_page


jurialmunkey:
Hey thanks for this! I'll run through a proper review soon. However, I noticed that you have declared encoding for every file which goes against PEP8 guidelines: 

> Files using ASCII (in Python 2) or UTF-8 (in Python 3) should not have an encoding declaration.

https://www.python.org/dev/peps/pep-0008/#source-file-encoding

My understanding of encoding declaration is that it doesn't change the encoding of strings in the code, all it does is tell the interpreter to read the code using that encoding. So if the file only contains ASCII characters then specifying an encoding declaration does nothing.

Ultimately harmless to specify but I _am_ trying to keep my code PEP8 compliant so thought I should point that out. 

Only place it really should be used is in discover.py due to the string literals for languages that have UTF-8 characters. There's actually a few that I need to clean-up which are either legacy leftovers or have snuck in due to copy-paste duplication.

dplim87:
Agreed. I just noticed that, at least in py2, using format with mixed Unicode and str types is where issues typically arise, because of py2 default encoding of ascii. A '{}{}'.format() call in a file with no encoding set would default to using ASCII in py2.

resources/lib/addon/parser.py
```diff
@@ -42,30 +62,30 @@ def try_type(value, output=None):
...
+        return string.encode(encoding, errors)
     except Exception:
         return string
 
 
 def parse_paramstring(paramstring):
     """ helper to assist with difference in urllib modules in PY2/3 """
     params = dict()
-    paramstring = paramstring.replace('&amp;', '&')  # Just in case xml string
+    paramstring = try_encode(paramstring).replace('&amp;', '&')  # Just in case xml string
```
jurialmunkey, line 87: Is this a typo? Wouldn't you want _decode_ for `str.replace()` and `str.split()` ? Also, pretty sure we want to return decoded params.

jurialmunkey (changes requested):
A few minor comments here and there but otherwise looks good!  😃 

Once the requested changes are in, I'll run it through some test cases and merge.

resources/lib/addon/parser.py
dplim87: not a typo, it's to utf-8 encode the string for the calls to unquote_plus

resources/lib/addon/parser.py
```diff
@@ -1,10 +1,30 @@
...
+        _query = dict()
+        for k,v in query.items():
+            _query[try_encode(k)] = try_encode(v)
+        return _urlencode(_query, **kwargs)
+
+    from urllib import unquote_plus as _unquote_plus
+    def unquote_plus(s):
+        if isinstance(s, unicode):
+            return _unquote_plus(try_encode(s))
+        else:
```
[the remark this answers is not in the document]
dplim87: I had originally left out the else: statement, but added it back for readability

resources/lib/addon/parser.py
```diff
@@ -42,30 +62,30 @@ def try_type(value, output=None):
...
 
 
-def try_decode(string, encoding='utf-8', errors=None):
+def try_decode(string, encoding='utf-8', errors='strict'):
     """helper to decode strings for PY 2 """
     if sys.version_info.major == 3:
         return string
     try:
-        return string.decode(encoding, errors) if errors else string.decode(encoding)
+        return string.decode(encoding, errors)
```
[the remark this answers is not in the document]
dplim87: I had figured set the default which is 'strict', and allow the caller to change if desired

resources/lib/addon/parser.py
dplim87: but after reviewing, these shims aren't needed... i swore unquote and unquote_plus threw errors when handed unicode values on py2, but alas they do not

resources/lib/addon/parser.py
```diff
@@ -74,19 +93,15 @@ def parse_paramstring(paramstring):
     return params
 
 
-def urlencode_params(*args, **kwargs):
+def urlencode_params(query, **kwargs):
```
dplim87, line 96: Perhaps remove this altogether and import urlencode from this module, which will resolve to either the shim or direct call?

from resources.lib.addon.parser import urlencode

resources/lib/addon/parser.py
jurialmunkey: > not a typo, it's to utf-8 encode the string for the calls to unquote_plus

In PY2 `encode()` transforms a unicode string `u'foobar'` into bytes `b'foobar'`. So in this method you are expecting the paramstring to be unicode. However, elsewhere in the container router you removed decoding from the paramstring so you will be getting bytes here no unicode.

I'd think that you'd want to decode the paramstring to unicode as early as possible (i.e. in container router) and then you don't need to encode/decode again because you're already doing try_encode in the shims.

resources/lib/addon/parser.py
jurialmunkey: I did have the same thought at first myself - my only concern was that for PY3 we directly import the module. But now that I think about it, it will be in the namespace since we import _from_ so it should work fine.

PR #518, opened 2021-06-24
setprop new additions: emmy_wins and emmy_nominations

inb4after:
resources/lib/monitor/common.py


jurialmunkey:
Haha of course. That's what I get for not actually testing something!

Thanks 😁

PR #610, opened 2021-12-31
Add new calendar episode property

roidy:
Add 'days_until_aired' property to episodes in the calendar.

This allows you to easily create labels like "Airs in 5 days"

Example
```
<variable name="AirsIn">
    <value condition="String.IsEqual(ListItem.Property(days_until_aired),0)">Airs today</value>
    <value condition="String.IsEqual(ListItem.Property(days_until_aired),1)">Airs tomorrow</value>
    <value>$INFO[ListItem.Property(days_until_aired),Airs in , days]</value>
</variable>
```

### DIFF
--- a/resources/lib/trakt/progress.py
+++ b/resources/lib/trakt/progress.py
@@ -244,8 +244,10 @@ class _TraktProgress():
             'air_time': get_region_date(air_date, 'time'),
             'air_day': air_date.strftime('%A'),
             'air_day_short': air_date.strftime('%a'),
-            'air_date_short': air_date.strftime('%d %b'),
-            'days_until_aired': str((air_date.date() - get_datetime_today().date()).days)}
+            'air_date_short': air_date.strftime('%d %b')}
+        days_to_air = (air_date.date() - get_datetime_today().date()).days
+        dtaproperty = 'days_from_aired' if days_to_air < 0 else 'days_until_aired'
+        item['infoproperties'][dtaproperty] = str(abs(days_to_air))
         item['unique_ids'] = {u'tvshow.{}'.format(k): v for k, v in i.get('show', {}).get('ids', {}).items()}
         item['params'] = {
             'info': 'details',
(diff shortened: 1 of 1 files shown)


jurialmunkey:
I added a few adjustments to deal with negative values: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/pull/610/commits/6ac15c6b92b3ee5e87ababb7eb77d7e8930a7126

If there's a negative value, I assign to `ListItem.Property(days_from_aired)` instead and use `abs()` to transform to a positive integer. That way you can check `!String.IsEmpty(ListItem.Property(days_from_aired))` to see if it aired in the past:

```xml
<variable name="AirsIn">
    <value condition="String.IsEqual(ListItem.Property(days_until_aired),0)">Airs today</value>
    <value condition="String.IsEqual(ListItem.Property(days_until_aired),1)">Airs tomorrow</value>
    <value condition="String.IsEqual(ListItem.Property(days_from_aired),1)">Aired yesterday</value>
    <value condition="!String.IsEmpty(ListItem.Property(days_from_aired))">$INFO[ListItem.Property(days_from_aired),Aired , days ago]</value>
    <value>$INFO[ListItem.Property(days_until_aired),Airs in , days]</value>
</variable>
```

I wasn't quite sure what to call the other property - days_from_aired was the best I could come up with. Does that make sense to you? Or can you think of a better way to approach the negative values issue? It feels a bit off having two properties but I can't think of a better way.


FYI - If you need to reset the cache to test, you can do that from Expert Settings. Restart Kodi after clearing the cache.
`TMDbHelper Settings > Expert > Delete Cache > Delete Trakt Cache`

roidy:
>I wasn't quite sure what to call the other property - days_from_aired was the best I could come up with. Does that make sense to you? Or can you think of a better way to approach the negative values issue? It feels a bit off having two properties but I can't think of a better way.

I hadn't even considered negative values 🤦‍♂️ But your solution seems perfectly fine. I've tested the changes and with the new variable it work great.

PR #636, opened 2022-02-05
Fix unwatched episodes counts

matke-84:
I think this fixes the problem I wrote about that there are no unwatched episodes in the entire tvshow as well as the lack info about a number of seasons.

### DIFF
--- a/resources/lib/container/router.py
+++ b/resources/lib/container/router.py
@@ -301,7 +301,7 @@ class Container(TMDbLists, BaseDirLists, SearchLists, UserDiscoverLists, TraktLi
                 season=li.infolabels.get('season'),
                 episode=li.infolabels.get('episode')) or 0
         if li.infolabels.get('mediatype') == 'tvshow':
-            li.infolabels['episode'] = self.trakt_api.get_episodes_airedcount(
+            li.infolabels['episodes'] = self.trakt_api.get_episodes_airedcount(
                 id_type='tmdb',
                 unique_id=try_int(li.unique_ids.get('tmdb')))
             return self.trakt_api.get_episodes_watchcount(


jurialmunkey:
plural `episodes` is not a valid infolabel name. Valid infolabels are defined here: https://alwinesch.github.io/group__python__xbmcgui__listitem.html#ga0f1e91e1d5aa61d8dd0eac90e8edbf18

The correct infolabel for episode count is `episode`

matke-84:
But this works. Have you tried it?

jurialmunkey:
> But this works. Have you tried it?

It works unintentionally  but it doesn't fix the issue. 

All it does is dump a `None` return from Trakt air count into a non-existent infolabel so that it doesn't overwrite the original data from TMDb. The underlying scenario of the None return from Trakt needs to be managed.

jurialmunkey:
FYI fixed here - https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/4506dd0b92576487238219c97102256420bee0fa

jurialmunkey:
Basically I added a check for Trakt returning None
```python
if li.infolabels.get('mediatype') == 'tvshow':
    air_count = self.trakt_api.get_episodes_airedcount(
        id_type='tmdb',
        unique_id=try_int(li.unique_ids.get('tmdb')))
    if not air_count:
        return
    li.infolabels['episode'] = air_count
```

This way if Trakt doesn't return an airedcount we don't retrieve the watchedcount and just return the original TMDb value for totalepisodes.
 
Reason we need airedcount from Trakt is because sometimes Trakt has different episode counts - usually because of "specials" seasons being different - and so if we compare Trakt watchedepisodes to TMDb totalepisodes we can end up with a wrong number.

For example, consider this scenario: 

- TMDb and Trakt have a show where S1 = 10 episodes. 
- Trakt also lists 5 specials but TMDb doesn't. 
- You watch those 5 specials. 
- You watch 5 episodes of season 1. 
- `TMDB totalepisodes (10) - Trakt watchedepisodes (5 episodes + 5 specials) = 0` so show is marked watched. Even worse, after actually watching all of season 1 the show has a count of `-5` unwatched.

Issue #632, opened 2022-01-27
Artwork quality


ambulancePilot:
Hey there. I am not sure if this is the right way to go about this. But I was wondering what quality artwork this addon grabs from TMDB and fanart.tv? And if there would be any way to set a preference to highest quality artwork? Maybe I am off base here, so please delete if not appropriate. I love the addon by the way. Great work. I use it daily and have never been disappointed!

jurialmunkey:
Fanart from both TMDb and fanart.tv are in the original source image format (i.e. highest quality). Posters from Fanart.tv are also in the original format (i.e. highest). 

Posters from TMDb are w500 (so around 500x735 depending on the poster's aspect ratio). I don't get originals because the default Kodi texture setting will downsample non-fanart images above 720p - i.e. anything outside of 1280x720 box that isn't 16:9 artwork gets scaled down anyway, so there's no point getting high res posters. 
https://kodi.wiki/view/Artwork/Cache#Texture_optimization

Also, contrary to intuition, higher res source images for posters can end up looking worse because Kodi is actually better at upscaling than it is at downscaling.

jurialmunkey:
More detailed history of bad image downscaling in Kodi is available in this thread:
https://forum.kodi.tv/showthread.php?tid=200401

It's not as bad as it used to be, particular into the texture cache itself but it's still a problem for "on-the-fly" downscaling done by the skinning engine -- so you really won't have much advantage getting higher res posters and most likely will end up with it looking worse.

ambulancePilot:
Thank you so much for the detailed answer! That's very helpful to know. Thanks again for the amazing work.

Further to the downscaling issue, I know exactly what you mean. For whatever reason, software finds it easier to upscale than downscale. I've seen super high resolution images get downscaled and look pixelated as a result.

PR #637, opened 2022-02-05
new option in artwork quality setting for TMDb

matke-84:
I added a few new options. Now we have:

Original
Uses highest quality available.
Prefers fanart.tv artwork (if enabled and available)

Very High
Uses W1280 fanart from TMDb
Prefers fanart.tv for all other artwork types (if enabled and available)

High
Uses W1280 fanart and W780 posters from TMDb
Prefers fanart.tv for all other artwork types (if enabled and available)

Medium
Uses W1280 fanart and W500 posters from TMDb
Prefers fanart.tv for all other artwork types (if enabled and available)

Low
Uses W1280 fanart and W342 posters from TMDb
Prefers fanart.tv for all other artwork types (if enabled and available)

very Low
Uses W780 fanart and W342 posters from TMDb
Prefers fanart.tv for all other artwork types (if enabled and available)

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -2027,11 +2027,6 @@ msgid "Retrieve season specific artwork"
 msgstr ""
 
 #: /resources/settings.xml
-msgctxt "#32397"
-msgid "Original"
-msgstr ""
-
-#: /resources/settings.xml
 msgctxt "#32400"
 msgid "Play local Kodi library files"
 msgstr ""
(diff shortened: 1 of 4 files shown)


jurialmunkey:
There's no point getting posters above w500. See my comments here: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/issues/632#issuecomment-1022751044

w500 will be pretty close to 720p (~500x735). Anything above 720p gets downscaled by Kodi texture cache (except fanart where the limit is 1080p). So going higher is a waste of bandwidth. 

Even with modified texture cache advancedsettings to allow it, at w780 the poster is taller than 1080p (~780x1147). Even on a 4k display, that's more than half screen height and most poster displays (e.g. wall/row view) display smaller posters than that. Considering that Kodi is *far* [better at upscaling than downscaling](https://forum.kodi.tv/showthread.php?tid=200401) - then a mid-sized w500 poster is the optimal size.

I appreciate the PR I really do! It's just that there was a rationale behind why I chose those specific settings. All that will happen is that I'll end up with complaints about worse image quality and longer loading times because everyone will put it on "original" thinking it is the "best" setting.

Issue #688, opened 2022-04-08
Short day and date for last and next aired


roidy:
In the trakt calendar we get both

ListItem.Property(air_day_short)
ListItem.Property(air_date_short)

Wed 06 Apr

Could we also get this format for item detail last_aired and next_aired and the window property versions please

jurialmunkey:
Those should already be in kodi system short format by default.

jurialmunkey:
Sorry, I see what you're saying now - you want `%d %b` formatting not system shortdate which is usually `dd/mm/yyyy` or similar.

Added in latest `next_aired.short` (and same for last_aired) - you'll need to refresh item details for already cached items to see changes.

roidy:
Sorry to be a nuisance but I also need the short version of the day name.

In the trakt calendar 
ListItem.Property(air_day) - gives 'Monday'
ListItem.Property(air_day_short) - gives 'Mon'

I need the same for item details
ListItem.Properety(next_aired.day) - gives 'Monday'
ListItem.Property(next_aired.day.short) - There is no short version :(

Also could we get the day number without the leading zero %-d
It's easy to add the leading zero using a variable if you want it, but impossible to remove it.

Thanks

PR #689, opened 2022-04-10
Add day_short property

roidy:
Addition to #688


roidy:
Not sure how you want to handle property naming with dots . or underscores _ or nothing

next_aired.day_short
vs
next_aired.day.short
vs 
next_aired.dayshort

jurialmunkey:
Thank for this PR.

Unfortunately `%-d` to remove the leading zero doesn't work on all platforms, so it needs to be `%d` for cross platform compatibility. Also keeps it consistent with calendar `air_date_short` property which is `%d %b`

In terms of naming - I think `next_aired.day_short` as you have it now is probably best option since in the calendar I called it `air_day_short`. I really should've standardized the naming convention better!

Revert the `%-d` back to `%d` and I'll merge.

roidy:
No problem, reverted now.

Issue #712, opened 2022-05-01
[Feature] Custom date formatting


roidy:
There is a skin setting to set a custom date format:- Skin.SetString(TMDbHelper.Date.Format,%d %b %Y)
However it only seems to be used for Premiered_Custom

I propose that every date  property should be able to use the custom set TMDBHelper.Date.Format ie.

Last_Aired.Custom
Next_Aired.Custom

Thanks

jurialmunkey:
Hmm, still trying to think about what the best approach would be for this.

Do you need the custom formatting in the ListItem or is a Window Property for focused item good enough?

roidy:
Window Property for focused item would be perfectly fine.

jurialmunkey:
Window property is much easier to manage. Working in v4.7.29

roidy:
Works perfectly, thanks 👍

PR #713, opened 2022-05-01
Allow custom date formatting for dates. Enhancement #712

roidy:
Allow info property dates to use the custom date formatting TMDbHelper.Date.Format
Adds a `.custom` property to dates like Last_Aired, Next_Aired etc...


jurialmunkey:
This approach won't work because the mapped metadata is cached in the ItemBuilder cache. So when you change the format, any previously cached items won't update, which will be a particular issue when switching skins.

roidy:
Ok, I see. So the custom format needs to be applied after the date is retrieved from the cache but before it's written to the item's property.

PR #782, opened 2022-06-28
Create dispatch_to_repo.yml

drinfernoo:
This allows this repository to dispatch to [repository.jurialmunkey](https://github.com/jurialmunkey/repository.jurialmunkey) whenever a tagged release is created.

This workflow requires that a personal access token be setup, as described [here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-token), and that it be added as a repository secret (named `TOKEN`), as described [here](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository).

This works in tandem with https://github.com/jurialmunkey/repository.jurialmunkey/pull/29.

### DIFF
--- a/.github/workflows/dispatch_to_repo.yml
+++ b/.github/workflows/dispatch_to_repo.yml
@@ -1,0 +1,21 @@
+name: Dispatch to Repo
+on: 
+  push:
+    tags:
+      - "*"
+  workflow_dispatch:
+    
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        repo: ["jurialmunkey/repository.jurialmunkey"]
+    steps:
+      - name: Push to repo
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.TOKEN }}
+          repository: ${{ matrix.repo }}
+          event-type: update
+          client-payload: '{"ref": "${{ github.ref }}", "sha": "${{ github.sha }}", "module": "jurialmunkey/plugin.video.themoviedb.helper", "tag": "${{ github.ref_name }}"}'


jurialmunkey:
Awesome! Giving it a test spin now. Wasn't sure if I needed to add the token to both ends (i.e. plugin.video.themoviedb.helper and repository.jurialmunkey) so I just added it to both. 

Action has run on tag creation but hit a snag with one of the submodules:

```
Run git submodule update --init --recursive --checkout --remote -f
Submodule path 'leia/context.themoviedb.helper': checked out '629a88be0b228d649819fe9c58dde2c9c8e50b0d'
fatal: Unable to find refs/remotes/origin/leia revision in submodule path 'leia/plugin.video.themoviedb.helper'
Error: Process completed with exit code 128.
```

https://github.com/jurialmunkey/repository.jurialmunkey/runs/7104561384?check_suite_focus=true

jurialmunkey:
Okay I'm stumped why this fails for leia branch of TMDbH. I tried deleting the submodule folder and the .git refs to re-pull it from scratch but it is still complaining that ref doesn't exist

jurialmunkey:
That was a bit of a journey but I'm fairly sure I've got it working correctly now. Ultimately what I had to do was

1. Add `fetch-depth: 0` to checkout action in https://github.com/jurialmunkey/repository.jurialmunkey/blob/86e467b2754c05de01ad15a95886c9b9b13b535f/.github/workflows/update_modules.yml#L18
2. Delete the `leia/plugin.video.themoviedb.helper` path and submodule.
3. Switch `_repo_xml_generator.py` to be `_repo_generator.py` in https://github.com/jurialmunkey/repository.jurialmunkey/blob/86e467b2754c05de01ad15a95886c9b9b13b535f/.github/workflows/update_modules.yml#L32
4. Update **all** .gitmodules to point to their relevant branches in https://github.com/jurialmunkey/repository.jurialmunkey/blob/master/.gitmodules

(1) is to checkout all branches because the default is to checkout only main. 

(2) I have no idea why but for some reason git kept complaining that the ref for the leia branch to TMDbH didn't exist not matter what I tried. I ended up just deleting the entire leia submodule as it was the only one causing issues. The leia version is so out of date now and there's not going to be any more updates to it so it doesn't really need to be in the repo anyway.

(3) Was just a file name mismatch that needed to be corrected

(4) Okay this was a pain. I finally got things working and then realised that all the leia branches got updated to the matrix versions. Was just a case of going through and specifying branches for all of them and then rerunning (there's a bunch of ugly commits in the history where I was trying to get it fixed as quick as possible).

drinfernoo:
1. I hadn't seen anything about `fetch-depth`, glad you found that!
2. 😲
3. A typo I should've caught before I made the PR 🤦‍♀️
4. I knew that would need to be done eventually, but I didn't realize they all needed to be set from the start 😅 

I saw that a few updates went out... does this seem working to you?

jurialmunkey:
> I hadn't seen anything about fetch-depth, glad you found that!

Yeah it was driving me nuts! Finally bothered to go read the pages for the checkout action and it was basically first thing they mention.

> 😲

I think it was because when I was still trying to keep leia up to date I was cherry picking a bunch of commits and it got really messy when I switched to the matrix branch being default. Anyway, not a big deal as almost two years since I updated the leia branch of TMDbH.

> A typo I should've caught before I made the PR

Honestly I should've seen it too! I finally got it all to trigger properly and got really excited and then it failed on that of all things!

> I knew that would need to be done eventually, but I didn't realize they all needed to be set from the start 

Neither. I thought at least those others were up-to-date so it shouldn't matter for the test run since it was only TMDbH updating. Didn't really put two and two together until it was too late.

> I saw that a few updates went out... does this seem working to you?

Yeah, after the few hiccups at the start, it seems to be working great now! I also set up the action for AH2 and it looks like it triggered properly for that too which is just amazing!

Honestly, I'm super excited about it because coupled with gh cli it is super simple now to push a release. I'm even thinking about other things I can script into it e.g. link pushing AH2 releases with Kodi's TexturePacker tool so that the release ships with a compressed .xbt for better performance rather than raw pngs

Really appreciate you getting this setup because I would've never done it myself, but now that it has been done I can't believe I didn't look into setting it up sooner!

PR #889, opened 2022-10-01
Update Arabic Translation

psp2111-ADSLGATE:
Hi @jurialmunkey 

Quick unrelated question: Can you check if TVDB section is working with you?


jurialmunkey:
Thanks, and no TVDb is not because I forgot to write code to refresh the token once it expires. Will fix shortly.

PR #892, opened 2022-10-02
Updated german translations

peat80:
Updated, tweaked and completed german translations of tmdb helper


jurialmunkey:
Excellent, thanks!

Issue #897, opened 2022-10-05
[question] Open recommendation script directly from actor info


matke-84:
I tried to use the recommendation script directly from actro info but there are small problems. When I enter the actor info and run the recommendation script, everything works as it should. Then when I click on a movie to open the info, I get the result as in video I posted. It starts to open and then returns to the script. Also, when I want to exit everything, a black screen appears and then I have to click back again to exit. Is there a chance to use it in this way or must it be used exclusively from the recommended window. Thank you in advance for your answer and I apologize that in the last few days I may have been a bit boring with questions and requests. My intention is just to help make tmdb helper as good as possible.

https://user-images.githubusercontent.com/74268331/193998508-7acd6199-3f76-4884-919e-58d70a073552.mp4

jurialmunkey:
I don't really understand what you're trying to do.

Actors work fine in AH2 using recommendations.

matke-84:
No, no, everything works when you enter the recommended script in a regular way. That's all ok. I tried a slightly different way to enter. Just watch the video carefully.
Clicking on the name of the actor I open the info about the actor here.
`RunScript(plugin.video.themoviedb.helper,add_path=$INFO[Container($PARAM[id]).ListItem.FolderPath],call_auto=1190)`
Currently in this window I have info about the actor and I made buttons when they are in focus to open the lists cast in movies, cast in tv show... Since the list of 50 movies opens a little slower, I tested the click ondown to open the recommended info, like you do in ah2.
`Runscript(plugin.video.themoviedb.helper,recommendations=.........`
And it works. But then when I click on the movie, it starts to open the info dialog, but instead of opening it, it returns it to the script. When exiting after that I get stuck on a black screen and have to click back again to exit.
When I do all this from the recommended window, everything works as it should.

jurialmunkey:
You can't mix the two window management methods.

Unnecessary anyway. Use the oninfo recommendations action. You can see how to use it from the cast list in AH2 video osd.

matke-84:
Woooow. I didn't know this. That's it. The recommended script is full of secrets. :-)

One small suggestion, I don't even know if it's possible. When I call the oninfo action from the info dialog, when returning back from the recommended script, it takes me back to home instead of the info dialog. Is it possible to return me to the info dialog instead?

https://user-images.githubusercontent.com/74268331/194050415-df7b9660-299e-41b7-b99d-443c9ed6cc73.mp4

[...]

PR #900, opened 2022-10-07
Reduced waiting time at the exit from recommendation script

matke-84:
I already left a comment that there is some delay when exiting from recommended window. With this it is reduced. #897

### DIFF
--- a/resources/lib/window/recommendations.py
+++ b/resources/lib/window/recommendations.py
@@ -404,14 +404,14 @@ class WindowRecommendationsManager():
         with WindowProperty((PROP_ONCLOSED, 'True')):
             executebuiltin(builtin) if builtin and not after else None
             executebuiltin(f'Dialog.Close(movieinformation,true)')
-            self._mon.waitForAbort(1)
+            self._mon.waitForAbort(0.1)
             if not cond and xbmcgui.getCurrentWindowId() == self._window_id:
                 _win = xbmcgui.Window(self._window_id)
                 _win.close() if _win else None
             executebuiltin(builtin) if builtin and after else None
             for _gui, data in self._history:
                 del _gui
-            self._mon.waitForAbort(1)
+            self._mon.waitForAbort(0.1)
             get_property(PROP_HIDEINFO, clear_property=True)
             get_property(PROP_HIDERECS, clear_property=True)
             get_property(PROP_TMDBTYPE, clear_property=True)


jurialmunkey:
Oh, it's that delay you're talking about.

The delay is necessary to make sure that the info dialog has closed properly. If the info dialog is still active, it will block Kodi from closing the underlying custom window and you will get stuck on an empty screen.

The reason it needs to be 1 second is because sometimes the executebuiltin command will be delayed due to a WindowClose animation firing. Most of the time 0.1 will be fine, but you'll end up getting stuck with this shorter delay.

A different solution is needed here to ensure that the dialog has properly closed first.

matke-84:
Yes, we need a solution here. The delay is a bit annoying. In AH2, it's not so noticeable because you have animations, and in my skin (titan bingie mod) it goes without animation, so it's really noticeable.
I also tried to solve the other problem from that post, when I use 'oninfo' takes me back to home.
`executebuiltin('Action(Info)') if self._recommendations == 'oninfo' else None`
And this actually works fine until I open the recommended window from the oninfo dialog. Then it returns me to home again. Here, too, it would be nice to we find some an elegant solution.

jurialmunkey:
There is no good solution for the second. There is no way to retrieve the existing listitem from the info dialog in order to save it to the history. Simulating pressing info is not a good idea as it can very easily break if the timing is wrong.

You'll notice when back tracking via normal method that the last info dialog is not actually the one you opened if it was from the library - instead you get tmdbs online version of the same item (eg it will be same movie butmight have different artwork from TMDb instead of the library). That's because it's not possible to retrieve the listitem itself to reopen the dialog on - it can only save the ID and then reconstruct a listitem using online info.

matke-84:
Ok. Certainly, it should not be complicated more than necessary. Stable work is more important. I'm going to keep this way of opening the list for now (cast in movies, cast in tv shows...), I'm just going to reduce a little bit the number of contents that it opens from 50 to maybe 30 in order to speed up a little bit of opening. I hope that you will find a solution to somehow eliminate that annoying delay when exiting from recommendation script.

matke-84:
Any progress with this? 🙂

davinciicodes:
@matke-84, if this is really urgent, I would recommend you submit another PR with a solution adapted to your needs and @jurialmunkey will review.

matke-84:
There is probably a reason why @jurialmunkey  hasn't solved this problem yet, unless he forgot. 🙂
I hope that a solution will be found soon.

PR #965, opened 2022-12-30
Allow play_using parameter to refer to addon name

MoojMidge:
I would like to be able to call TMDb Helper players, without actually knowing the filename of the player json file, but instead based on the name of the plugin that is used as the plugin ID in the player file.

This PR is the simplest way I could figure out how to do that, and just searches the list of players to match the first player whose plugin ID is equal to the value of the play_using parameter, if a filename match is not already found.

### DIFF
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -17,7 +17,6 @@ from resources.lib.player.putils import get_players_from_file, make_playlist, ma
 from resources.lib.files.futils import get_json_filecache
 from resources.lib.addon.logger import kodi_log
 from threading import Thread
-from unicodedata import normalize as u_normalize
 
 
 def string_format_map(fmt, d):
@@ -163,7 +162,7 @@ class Players(object):
                     priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
                 player['is_provider'] = False    
             player['priority'] = priority
-            return priority, u_normalize('NFKD', player.get('name', '\uFFFF')).lower()
+            return priority, player.get('plugin', '\uFFFF').lower()
 
         players = sorted(self.players.items(), key=_set_priority)
         return players
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -174,10 +174,12 @@ class Players(object):
         if value:
             file = player_id
         else:
-            file, value = next(
-                ((file, value) for file, value in self.players.items() if value.get('plugin') == player_id),
-                (player_id, {})
-            )
+            for file, value in self.players.items():
+                if value.get('plugin') == player_id:
+                    break
+            else:
+                file = player_id
+                value = {}
         if mode in ['play_movie', 'play_episode']:
             name = get_localized(32061)
             is_folder = False
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -146,16 +146,17 @@ class Players(object):
 
     def _get_prioritised_players(self):
         try:
-            providers = self.details.infoproperties.get('providers', '')
+            providers = self.details.infoproperties.get('providers')
+            if providers:
+                providers = providers.split(' / ')
         except AttributeError:
-            providers = ''
-        providers = tuple(providers.split(' / '))
+            providers = None
 
-        def _set_priority(item, _providers=providers):
+        def _set_priority(item):
             file, player = item
-            player_provider = player.get('provider') or None
-            if player_provider in _providers:
-                priority = _providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
+            player_provider = providers and player.get('provider')
+            if player_provider and player_provider in providers:
+                priority = providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
                 player['is_provider'] = True
             else:
                 if player.get('is_provider', True):
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -158,7 +158,9 @@ class Players(object):
                 priority = _providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
                 player['is_provider'] = True
             else:
-                priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
+                if player.get('is_provider', True):
+                    priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
+                player['is_provider'] = False    
             player['priority'] = priority
             return priority
 
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -17,6 +17,7 @@ from resources.lib.player.putils import get_players_from_file, make_playlist, ma
 from resources.lib.files.futils import get_json_filecache
 from resources.lib.addon.logger import kodi_log
 from threading import Thread
+from unicodedata import normalize as u_normalize
 
 
 def string_format_map(fmt, d):
@@ -162,7 +163,7 @@ class Players(object):
                     priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
                 player['is_provider'] = False    
             player['priority'] = priority
-            return priority, player.get('name', '\uFFFF').lower()
+            return priority, u_normalize('NFKD', player.get('name', '\uFFFF')).lower()
 
         players = sorted(self.players.items(), key=_set_priority)
         return players
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -153,14 +153,14 @@ class Players(object):
 
         def _set_priority(item, _providers=providers):
             file, player = item
-            player_provider = player.get('provider')
+            player_provider = player.get('provider') or None
             if player_provider in _providers:
                 priority = _providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
                 player['is_provider'] = True
             else:
                 if player.get('is_provider', True):
                     priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
-                player['is_provider'] = False    
+                player['is_provider'] = False
             player['priority'] = priority
             return priority, player.get('plugin', '\uFFFF').lower()
 
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -162,7 +162,7 @@ class Players(object):
                     priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
                 player['is_provider'] = False    
             player['priority'] = priority
-            return priority
+            return priority, player.get('name', '\uFFFF').lower()
 
         players = sorted(self.players.items(), key=_set_priority)
         return players
--- a/resources/lib/player/players.py
+++ b/resources/lib/player/players.py
@@ -169,16 +169,15 @@ class Players(object):
                     return False  # Key didn't have a value so player fails assert check
         return True  # Player passed the assert check
 
-    def _get_built_player(self, file, mode, value=None):
-        value = value or self.players.get(file) or {}
-        if not value:
-            plugin_name = file
-            plugin_players = [
-                (file, player) for file, player in self.players.items()
-                if player['plugin'] == plugin_name
-            ]
-            if plugin_players:
-                file, value = plugin_players[0]
+    def _get_built_player(self, player_id, mode, value=None):
+        value = value or self.players.get(player_id)
+        if value:
+            file = player_id
+        else:
+            file, value = next(
+                ((file, value) for file, value in self.players.items() if value.get('plugin') == player_id),
+                (player_id, {})
+            )
         if mode in ['play_movie', 'play_episode']:
             name = get_localized(32061)
             is_folder = False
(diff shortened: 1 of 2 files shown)


jurialmunkey:
I like the idea and looks mostly good. 

However, a couple slightly more unusual but still likely test cases to consider:

1. User has multiple player files for the same plugin
2. The player selected doesn't have the required mode

For (1) the player priority should be considered.
For (2) also consider that some users might split movies and episodes modes for a plugin into separate player files for fallback reasons. So there might be a player available with that mode but it won't necessarily be the first result even if we sort by priority.

MoojMidge:
This definitely wasn't the most comprehensive change, just the simplest.

Will update PR to cover different priorities and modes.

Set to draft in the meantime.

jurialmunkey:
No worries. Honestly, was very close to merging originally as I didn't think of those cases immediately and it appear to work well.

It's only because I've got a couple of Australian TV streaming addons that have really inconsistent metadata where I've been thinking of cascading a few players and it made me think how this might affect that approach.

Also apologies in advance as the player code is a bit of a mess! Definitely something I would've done differently if starting over. It's a lot cleaner than it used to be but there's so much legacy backwards compatibility crap in there that's difficult to unpick.

MoojMidge:
Australian catchup TV plugins were my use case for this as well.

No need to apologise, like I said before no project or codebase is perfect, and the player functionality is a great option for managing video plugin playback. Saves me having to start from scratch when the hard work is already done.

In terms of legacy stuff, one thing I noticed was that `player` and `mode` arguments were passed through here when used as a script: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/c18bb313a96f28a827f62722ee7e7406d207dd05/resources/lib/script/method.py#L508-L510

But not when using a plugin url and passing through the `info=play` and `play_using` parameters:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/c18bb313a96f28a827f62722ee7e7406d207dd05/resources/lib/items/router.py#L25

Is this intended?

jurialmunkey:
Yeah thats intentional. Via the script, play_using= is transformed into player= because "player" is too generic to use as a routing keyword. And since it explicitly sets a player, then we also need mode which is set to "play" by default.

Mostly the script function is setup that way to make it a bit more user friendly and a bit less verbose - `play_using=playername` is much easier than `info=play, player=playername,mode=play`

Also allows for restricting exactly which methods are exposed by the script rather than simply reusing the plugin routing.

MoojMidge:
Ah OK. I was just looking at the wiki which had the script parameters and wondering why the same parameters weren't working for the plugin routing.

The last commit in this PR does change the plugin routing so that the previously required three parameters 
```
?info=play&player=[playerfile|playername|provider]&mode=[play|search]
```
Is reduced to only requiring the following (undefined mode defaults to using play method of selected player):
```
?player=[playerfile|playername|provider]
```
Also allows the following alternatives:
```
?play_using=[playerfile|playername|provider]

?play_using=[playerfile|playername|provider]&mode=[play|search]

?player=[playerfile|playername|provider]&mode=[play|search]
```

This was mostly because I wanted to be able to use the same parameters for the plugin path or the script call (if the plugin path resolving failed because of a plugin overriding normal playback).

jurialmunkey:
I wouldn't want to change the info routing on plugin path. Only info= should determine routing in the plugin path because then it means that there is no concern about different methods reusing keywords for params. 

It is better to change the script routing if anything but I don't really understand the need anyway. You already need to configure a paramstring for runplugin() differently from comma separated arguments for a runscript() command anyway, so just do the conversion at that same point on the other addon.

MoojMidge:
All good, I was just initially confused based on the wiki content, with the script parameters not doing anything when used for the plugin url.

Will remove the changes to the routing, easy enough to just use the `info`, `player`, and `mode` parameters, which will work for both the plugin and the script. Probably won't end up using the script call anyway.

resources/lib/player/players.py
```diff
@@ -145,17 +146,27 @@ def _set_external_ids(self, tmdb_type, tmdb_id, season, episode, details, requir
...
+            player_provider = player.get('provider')
+            if player_provider in _providers:
+                priority = _providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
+                player['is_provider'] = True
+            else:
+                if player.get('is_provider', True):
+                    priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
+                player['is_provider'] = False    
+            player['priority'] = priority
+            return priority, u_normalize('NFKD', player.get('name', '\uFFFF')).lower()
```
jurialmunkey: What's the purpose of normalising the name here? Why does the name matter in terms of sorting? 

Existing read order from filenames should be enough in case a plugin has two players with the same priority. That way label formatting like `[COLOR=blue]` in names doesn't influence sort order and there isn't this extra normalization process slowing things down potentially unnecessarily.

There shouldn't ever be more than one player per provider, so no need to sort those out, and otherwise priority should be up to the user to determine.

resources/lib/player/players.py
```diff
@@ -145,17 +146,27 @@ def _set_external_ids(self, tmdb_type, tmdb_id, season, episode, details, requir
...
+
+        players = sorted(self.players.items(), key=_set_priority)
```
jurialmunkey, line 168: I think expanding the `_set_provider_priority()` function to handle sorting unnecessarily complicates its purpose.

My suggestion would be to leave the _set_provider_priority function unmodified. Then add a new function to move sorting out of `_get_players_for_dialog` and up to `__init__`

e.g.
```python
self._set_provider_priority()
self.players_prioritised = self._get_sorted_players()

[...]

def _get_sorted_players(self):
    return sorted(self.players.items(), key=lambda i: int(i[1].get('priority') or 0) or PLAYERS_PRIORITY)
```


Ideally functions should be kept single purpose. Write a new function for new functionality. Keeps things readible, simple, and modular. (I know my legacy code often breaks this principle but I try to apply it to any new code being added).

_set_provider_priority had one purpose previously: if an item has providers listed, then it modifies provider players to have higher priority (e.g. "Stranger Things" is provided by Netflix so Netflix player get higher priority).

That's also the purpose of the try/except block wrapper around the `['providers']` key rather than using `.get()` -- skips unnecessary iteration and modification over the players when the item doesn't have any providers.

jurialmunkey (changes requested):

resources/lib/player/players.py
MoojMidge: The intent was to have players with the same priority sorted in an easily understandable manner (alphabetically) in the player select dialog.

The original filename sort may be off because of the "autogen" prefix, or simply because the file naming convention is not specified anywhere. From what I have seen the filename for players may the provider name, others may be the plugin name e.g. "plugin.xxx.yyy", others may have other various prefixes e.g. "direct".

I didn't consider that the player name may have BBCode in it, which would throw things off in much the same way as the differing file naming conventions would.

In terms of the normalisation, that was just to deal with a hypothetical scenario where the player names included non-ASCII characters, so that a player name starting with `Ä`, or any other Unicode character, was not sorted by default after all the ASCII names.

Perhaps an alternative that addresses all four issues - different file naming conventions, BBCode in player names, non-ASCII characters, and potentially slow normalisation - is to just use the plugin name as the second sort key, as that is shown as label2 in the dialog entries?

Or I can just remove the second sort key.

resources/lib/player/players.py
MoojMidge: The idea was to minimise the number of times the players were sorted and looped through.

If there were providers, the players were looped through. The players were then being sorted and then looped through again to get dialog entries.

And in order to select a player by plugin name or provider, via the player parameter, while also respecting player priority and mode, I needed to sort and loop through the players again.

The aim of this PR was to perform the selection by name/provider via the player parameter, and for that I need to update the priorities, sort by priority, then loop through the result. The fact that the provider priority is being updated was secondary to that, because the sorting with priority as the sort key was required anyway, which is why I removed the KeyError handling.

Perhaps _set_provider_priority can be renamed to _get_prioritised_players? It would save having to iterate over players twice in a row if providers were present.

resources/lib/player/players.py
MoojMidge: 9f8e7ec removes the normalisation and uses plugin name as a secondary sort key

resources/lib/player/players.py
MoojMidge: [9c4ffee](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/pull/965/commits/9c4ffeecc14336c69e20cadf1e2338332facf99d) renames _set_provider_priority to _get_prioritised_players, but if you prefer, I can split this up into two different loops - setting priority in _set_provider_priority, and sorting by priority in _get_prioritised_players.

resources/lib/player/players.py
jurialmunkey: Probably not worth the trouble trying to come up with foolproof solution. I think on the balance, plugin name is a good enough fallback.

Easy enough for users to adjust priority according to their own preferences via the customisation dialog.

resources/lib/player/players.py
jurialmunkey: Oh, of course, I see what's happening here now! You've skipped the first for loop by making it part of sorted() key function.  

I didn't pick that up at first, so was still thinking there would be two loops either way. Hadn't thought of using sorted() as a substitute for loop before. That's actually really clever.

In that case, I'm definitely happy with that change. Cutting out an unnecessary loop beats slightly easier to read code.

Thanks for persisting with me here! Your code is definitely 100% an improvement and was just my idiot brain misunderstanding exactly what it was doing.

resources/lib/player/players.py
```diff
@@ -144,18 +144,28 @@ def _set_external_ids(self, tmdb_type, tmdb_id, season, episode, details, requir
...
+
+        def _set_priority(item, _providers=providers):
+            file, player = item
+            player_provider = player.get('provider')
+            if player_provider in _providers:
+                priority = _providers.index(player_provider) + 1  # Add 1 because sorted() puts 0 index last
+                player['is_provider'] = True
+            else:
+                if player.get('is_provider', True):
+                    priority = player.get('priority', PLAYERS_PRIORITY) + 100  # Increase priority baseline by 100 to prevent other players displaying above providers
```
jurialmunkey, line 163: Shouldn't this be `if not player.get('is_provider')` ?

If the player is **not** a provider then we want to bump its priority value by 100 to force it further down.

resources/lib/player/players.py
```diff
@@ -144,18 +144,28 @@ def _set_external_ids(self, tmdb_type, tmdb_id, season, episode, details, requir
...
-                v['priority'] = providers.index(v['provider']) + 1  # Add 1 because sorted() puts 0 index last
-                v['is_provider'] = True
-        except (KeyError, AttributeError):
-            pass
+    def _get_prioritised_players(self):
+        try:
+            providers = self.details.infoproperties.get('providers', '')
+        except AttributeError:
+            providers = ''
+        providers = tuple(providers.split(' / '))
```
jurialmunkey: What's the advantage of converting the list from .split into a tuple?

Also note that `''.split(' / ')` will give `['']` which potentially will cause false positive matches - e.g. if a player file has `provider: ""` rather than not including it. See my comment below for a simple workaround.

resources/lib/player/players.py
```diff
@@ -144,18 +144,28 @@ def _set_external_ids(self, tmdb_type, tmdb_id, season, episode, details, requir
...
+
+        def _set_priority(item, _providers=providers):
+            file, player = item
+            player_provider = player.get('provider')
```
jurialmunkey: I think might be a good idea here to make sure if a blank string is returned it gets converted to None type:
`player_provider = player.get('provider') or None`

That way we avoid `player_provider = ''` matching `_providers = ''.split(' / ')`

jurialmunkey (changes requested):
Thanks for persisting with me here! Above changes are great.

I think we're pretty close to merging. A couple of very minor comments below for a few small things I realised while testing.

resources/lib/player/players.py
MoojMidge: First a bit of context. I am currently just using this by creating plugin urls, in which case a new instance of the Players class is created everytime it is needed. In this scenario the sorting is only done once in the life of the Players instance.

However I was thinking about potentially importing and using the players module to create a longer lived Players instance.

In this scenario, the sorting may need to be done more than once (everytime a new listitem was required to be played using the Players instance).

If the sorting is done more than once, `not player.get('is_provider')` means that the priority gets bumped everytime a provider is not matched. However it only needs to have it's priority changed if it has never been sorted before `('is_provider' not in player)` or if it has been sorted and was previously a priority provider player `(player['is_provider'] == True)`. If it has been sorted and already had the priority changed `(player['is_provider'] == False)` then the priority shouldn't get bumped down again.

That is what the `player.get('is_provider', True)` condition does - it checks for `is_provider` being `True` (already sorted and previously increased priority) or not present (first time being sorted).

resources/lib/player/players.py
MoojMidge: Computer says no.

I use pylint and it complains about using a mutable default value for the `_providers` parameter of `_set_priority`.

A tuple being equivalent to an immutable list makes it happy, but it doesn't bother me either way, especially as there is no modification happening.

resources/lib/player/players.py
MoojMidge: Yep, makes sense. Updated in [386bb11](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/pull/965/commits/386bb11f926b2ccb24b5d4b11a8d8bdcbc270eaf)

resources/lib/player/players.py
MoojMidge: All good. Trying to review someone else's code isn't always easy, and you need to be comfortable with the changes. Thanks for taking the time to review.

resources/lib/player/players.py
jurialmunkey: Ah okay gotcha. 

I was only thinking in terms of single use since it has to be that way for plugin instances, so I couldn't quite figure out the purpose of writing the condition backwards. Makes way more sense now with the context of reusing and resorting.

resources/lib/player/players.py
jurialmunkey: Ah I see. Yeah dict as default arg is bad.

More to the point though: there's not really any need for the `_providers=` arg at all.

`_set_priority()` is an inner function of `_get_prioritised_players` so it has access to its variable scope. You may as well skip defining `_providers` entirely and just use `providers` directly instead -- unless there's something I missing here?

MoojMidge: No, not missing anything, it's just a (bad) habit I developed trying to speed up nested or looped function calls.

In theory it makes the variable lookup faster, without having to do the extra dereference, but in practice only makes a difference in much longer loops.

[250c735](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/pull/965/commits/250c7353c1e4438afdc9fb6c03097b7890d7f7a5) cleans things up a bit, and also avoids the list membership check if either `providers` or `player_provider` is not defined or an empty string.

jurialmunkey: Ah I see. Yeah might be a bit of premature optimisation there. I doubt we'd get anywhere close to the size of loops needed for it to be noticable (I have other concerns if someone has that many players installed!).

Side note, had never thought of using `and` like that for variable assignment before. That's really clever. Will need to remember that for the future.

I'm happy with everything here and will merge/release. I'm sure I'll hear quick enough if we've broken something but all seems to be working well with my tests. Really great additions and also very keen for these corresponding UpNext improvements.

jurialmunkey (approved):

jurialmunkey:
Merged and also backported to Matrix version.

Thanks again for these changes and persisting with me. Really great additions and looking forward to the corresponding changes planned for UpNext.

MoojMidge:
All good, thanks for taking the time to consider and review.

PR #981, opened 2023-01-22
Update player template for Slyguy plugins

MoojMidge:
These plugins have an option to flatten single season shows, which is not currently handled by the players created from the template

### DIFF
--- a/resources/templates/plugins_slyguy.json
+++ b/resources/templates/plugins_slyguy.json
@@ -12,13 +12,14 @@
                           },
     "play_movie"        : [
                             "plugin://STR_PLUGINID/?_=search&query={title_url}",
-                            {"strict": "true", "title": "(?i)^(\\[.*\\])?.*{title}.*(\\[.*\\])?$", "year": "{year}", "dialog": "true"}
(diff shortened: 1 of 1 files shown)


jurialmunkey:
LGTM.

BTW - when `strict` is specified there's no real difference between `'dialog': 'true'` and `'dialog': 'auto'`. Strict only returns a list for the dialog if there's more than one item matching the rule and auto only autoselects if the dialog list has a single item -- there's no scenario where auto can come into play with strict enabled because a single match is selected before it gets to the dialog.

Basically:

- strict=true + dialog=true/auto
  - 0 matches = fail
  - 1 match = select
  - 2 (or more) matches = dialog of matches
- strict=false + dialog=true 
  - 1 (or more) matches = select first match
  - 0 matches = dialog of entire folder 
- strict=false + dialog=auto
  - 1 (or more) matches = select first match
  - 0 matches + 1 item in folder = select item
  - 0 matches + 2 (or more) items in folder = dialog

MoojMidge:
Thanks for the clarification.

PR #983, opened 2023-01-23
Add xbmc.python.module extension point

MoojMidge:
I am currently using TMDB Helper in a way that is not fully supported via the existing script and plugin interfaces.

Rather than create PRs for functionality that really has no purpose except for my specific use case, I thought it would be better to allow TMDB Helper to be imported, so that I can expand on the functionality in my own plugin first. T

I can then raise any relevant PRs in the future for your consideration, if there is anything that may be more broadly useful.

Currently I just override some player configuration options (disable dialogs, force strict mode, disable playlist creation, allow playlist queueing without actual playback) to streamline third party plugin integration. See [here](https://github.com/MoojMidge/service.upnext/blob/master/resources/lib/tmdb_helper.py) for details if you are interested.


jurialmunkey:
I've thought about this before and have been tempted to do it in the past.

My main concern would be other plugins using my API keys. I know that technically they could steal them anyway but I'd want to make it at least not the default behaviour -- other end should need to supply a key for their project so it doesn't artificially inflate my usage stats.

Other concern is that it potentially exposes a user's accounts (e.g. Trakt, MDBList) info to other add-ons that the user didn't explicitly authorise. I'm not paritcularly concerned about malicious behaviour (a malicious addon can do far worse than steal Trakt data anyway), but I'd want to make sure to respect user choice. 

e.g. If I'm a user authorising Trakt for TMDbHelper, I'd be thinking that only TMDbHelper has access to it, not also any other application which wants to import TMDbHelper as a module. If some other application is gaining access via TMDbHelper then that authorisation should be made explicit to the user (at least as a one off agreement on first new access) and there should be a way to revoke it.

I don't think any of these classes you expose here have Trakt access but I'd need to double check to make sure. 

Overall I'd be keen to add this functionality as allow other addons access via the same centralised caching database could potentially be a huge benefit - just need to think through the implications a bit first as it does expose quite a bit, and would also need to address the API key issue.

MoojMidge:
Right, fair point(s) that I didn't think about at all.

I've come up with a provisional method that might address these issues. Will do some testing to see if it is a practical solution. Changed to draft in the meantime.

MoojMidge:
So I have updated this PR to include a check for permissions that are set prior to importing any modules.

To change the permissions required for any module, the following bit of code should be added to the top of the module:
```
import sys
from resources.lib.addon.consts import PERMISSIONS
if PERMISSIONS('general', 'tmdb') - getattr(sys.modules.get('themoviedb_helper'), '__permissions__', PERMISSIONS('all')):
    raise ImportError('Access denied')
```
For this example module `PERMISSIONS('general', 'tmdb')` are required to directly or indirectly import this module using `themoviedb_helper` as a central entry point for any 3rd party import. The equivalent of `import resources.lib.player.players` for a 3rd party addon would be `import themoviedb_helper.player.player`

The permissions granted are currently hard-coded to `PERMISSIONS('general')` with other options being any combination of the following as parameters to `PERMISSIONS` function.
```
    'general'
    'fanarttv'
    'mdblist'
    'omdb' 
    'tmdb' 
    'trakt' 
    'tvdb' 
```
With the hard-coded `PERMISSIONS('general')`, any attempt to import `api.tmdb.api` (which requires `PERMISSIONS('general', 'tmdb')`),  either directly or indirectly via an existing module, will fail when using a 3rd party addon.

Existing import functionality for the plugin, script and service remain unchanged.

Note that this method is not secure at all, as any 3rd party addon could still bypass all of this by editing the `addon.xml`, modifying `sys.path`, or manually importing modules based on a file path, etc., but for "nomal" usage seems to work correctly.

It also requires that every package/folder in TMDB Helper have a `__init__.py` file with `__all__` set to a mutable list of all modules available to import from that package.

If this approach is on the right track I will add the required permissions to the modules that expose the other various APIs being used, and add a new settings page where users can enable/disable the granted permissions.

jurialmunkey:
I'm going to stop you here, partly because at the moment I don't want to introduce a dynamic import approach, but more importantly because I don't think these changes are addressing my main concerns:

1. If a 3rd party addon imports TMDb() then the **_addon author_** should supply their own api_key rather than use mine
2. If a 3rd party addon imports TraktAPI (or MDblist etc) then the **_end user_** should be prompted to authorise the third party addon to allow access to their Trakt account.

The second point isn't that important at the moment because your original commit wasn't exposing Trakt. Yes, the Players class does use Trakt, but only for publicly available data (unique IDs) and not private user data.

However, the point about API project keys needs to be addressed. Currently TMDb API key is hard coded into the class here:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/39ae1b05bf0da54a839700a1d78442dded96e5e5/resources/lib/api/tmdb/api.py#L27-L31

This should be moved out to a variable in the constants. Internally TMDbHelper should use the constant for the API key value, but when imported by another addon that value should be blank -- i.e. third party addons need to supply their own API project keys.

I'm not fussed about a "secure" approach (my key is plain text afterall) and I don't care about restricting which addons have access to the module, I just don't want other addons using my API key automatically when they do import the module.

BTW - I'd consider Up Next using Players as an extension of TMDbHelper functionality, so I'm not fussed if you or similar projects use my key purely for Players access. My concern is unrelated addons using my API key instead of obtaining their own key. TMDb provides free keys because the point is that each project should have their own key to give better representation of usage etc.

MoojMidge:
> * If a 3rd party addon imports TMDb() then the **_addon author_** should supply their own api_key rather than use mine
> * If a 3rd party addon imports TraktAPI (or MDblist etc) then the **_end user_** should be prompted to authorise the third party addon to allow access to their Trakt account.

The eventual aim is to achieve just this, but some preliminary work is required to get it working.

> This should be moved out to a variable in the constants. Internally TMDbHelper should use the constant for the API key value, but when imported by another addon that value should be blank -- i.e. third party addons need to supply their own API project keys.

> I'm going to stop you here, partly because at the moment I don't want to introduce a dynamic import approach,

I'm not sure it will be possible to address the main issues without having some kind of dynamic import system. The original approach I used was to just hard-code the modules I wanted to use, but this would require changes if another 3rd party addon required something different.

Additionally, and more importantly, simply moving the API keys to a separate module, wouldn't work because a mechanism is required to prevent that separate file from being imported.

For example if `api.TMDB.api` was modified to remove the hard-coded API key, and instead used something like the following `from resources.lib.addon.api_keys import TMDB_API_KEY`, where `TMDB_API_KEY` was used if available when creating a new instance of the `TMDb` class, then whenever `api.TMDB.api` was made available to use as a module, and was imported by a 3rd party addon, the `TMDB_API_KEY` import would still occur as normal, as that is just how the standard Python import mechanism works.

A non-hardcoded import mechanism is required to be able to stop the import of `TMDB_API_KEY` for whatever other internal or external module may import it, but only if it is being imported through the Kodi Python module endpoint. That is what I have tried to implement.

Another option would be to have wrapper modules that glue the API modules together with the API keys, e.g. something like the following that would need to be used instead of importing `api.TMDB.api` directly within TMDB Helper.
```
from resources.lib.addon.api_keys import TMDB_API_KEY
import resources.lib.api.TMDB.api
```
Then `api.TMDB.api` can be hard-coded to be exposed through the Kodi Python module endpoint. This just seemed ugly to me and would have required a lot of changes to existing TMDB Helper code simply to facilitate the use of the wrapper modules.

jurialmunkey:
> The original approach I used was to just hard-code the modules I wanted to use, but this would require changes if another 3rd party addon required something different.

I'm not worried about other addons for now. Main goal at the moment is allowing Up Next access to what it needs without leaving access to TMDb via my API key wide open.

Exposing all the modules now is going to paint me into a corner where I can't change functions or refactor without risking breaking other people's plugins. I'm happy enough with where the TMDb api module is at to have it exposed and be confident I'm wont need to make breaking changes, but a lot of the other modules simply aren't there yet.

Eventually I want to split the plugin up into properly separate module addons anyway. I've already made some movement towards this by splitting out some common functions into script.module.tmdbhelper. I'd rather take this approach of incremementally splitting parts off for third party usage than jumping the gun and making everything available now.

> Additionally, and more importantly, simply moving the API keys to a separate module, wouldn't work because a mechanism is required to prevent that separate file from being imported.

Sorry, bit of a misunderstanding here. I wasn't suggesting that as a solution. I only meant to point out that it is currently hardcoded and needs a way to not be hardcoded. 

Simple idea I had now that I think *might* work: move API key to a class attribute and then override it in extension module

Basically TMDb class `_init_` gets changed like this: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/6b81901aa09aa2e2f3b6f680dc33be9d032d85b0
```python
class TMDb(RequestAPI):

    _api_key = 'a07324c669cac4d96789197134ce272b'

    def __init__(
            self,
            api_key=None,
            language=get_language(),
            mpaa_prefix=get_mpaa_prefix()):
        super(TMDb, self).__init__(
            req_api_name='TMDb',
            req_api_url=API_URL,
            req_api_key=f'api_key={api_key or self._api_key}')
```

Then in the extension point just override the `_api_key` class attribute to be None

```python
from resources.lib.api.tmdb.api import TMDb as TMDbBase

class TMDb(TMDbBase):
    _api_key = None

__all__ = (
    'TMDb',
)
```

Haven't tested in terms of importing as a module but it worked as expected when I did it internally.

MoojMidge:
> I'm not worried about other addons for now. Main goal at the moment is allowing Up Next access to what it needs without leaving access to TMDb via my API key wide open.

Appreciate you considering this use case.

> Exposing all the modules now is going to paint me into a corner where I can't change functions or refactor without risking breaking other people's plugins. I'm happy enough with where the TMDb api module is at to have it exposed and be confident I'm wont need to make breaking changes, but a lot of the other modules simply aren't there yet.

The dynamic importing is governed by the `__init__.py` for each package and/or the permissions that are set for each module. The current commits expose everything, but that is easily remedied.

That been said, I'm guessing this would be the first such use of TMDB Helper in this way. It provides a lot more flexibility compared to the alternatives of the script/plugin endpoints, and the pros greatly exceed the cons of things breaking. Maybe if this does get across the line, it just needs to be labelled as experimental and subject to frequent changes/breakages. It's not like you owe anyone any service guarantee anyway.

As for why I am persisting with the dynamic importing, even in this proof of concept implementation - I just can't see another way to do this without exposing your API keys and that the end-users keys/tokens.

> Simple idea I had now that I think _might_ work: move API key to a class attribute and then override it in extension module

Agree that separating the API keys will be required, but I am finding that overriding it properly is proving much more complicated.

Have made the first step towards doing this in the most recent bunch of commits. This will allow any module to be imported, if allowed in the package `__init__.py`, with permissions able to be set for whole modules or individual module variable/attributes/methods. All TMDB Helper API keys are blocked from being exposed via the module endpoint. User API keys/tokens can be exposed if appropriate permissions are set. This would need to be added to the addon settings, for a user to enable their API keys/tokens to be exposed.

> Haven't tested in terms of importing as a module but it worked as expected when I did it internally.

This works when the `TMDb` class is imported and used directly, however every other instance of `TMDb` will still use the original class with the original `api_key` class attribute. For example in `player.players.Players`, `TMDb` is used indirectly in a number of different places because `player.players` imports a bunch of functions from `player.details` that use their own `TMDb` instances. Same with the use of `ItemBuilder` and a bunch of other modules/classes that get pulled along for the ride when using `player.players.Players`.

Short of re-architecting the structure of all those modules so that the API modules can be passed as parameters to a class constructor, or something similar, to achieve a suitable level of indirection where each module doesn't end up creating their own API instances, then I don't think it will be possible to prevent the API keys from leaking in this way.

jurialmunkey:
> This works when the TMDb class is imported and used directly, however every other instance of TMDb will still use the original class with the original api_key class attribute.

Well that's why it'd only be suitable with a restricted set of modules being made available.

Like I said before, I'm not concerned about my API key being used for Player functionality. Having access to the module doesn't expose much more than what is already available via RunScript methods. It just gives more direct access. I'd consider that an extension of base functionality.

The difference with having direct access to the TMDb api class is that it allows for making any request to TMDb. I don't want someone building a plugin that imports my TMDb module and makes 100k daily requests via my API key.

> it just needs to be labelled as experimental and subject to frequent changes/breakages. It's not like you owe anyone any service guarantee anyway.

If only people were that reasonable, sensible, and unentitled...

What happens in reality is people jump on the experimental features and implement early even if I say it isn't ready, then they complain that there's no documentation or that I broke all their hard work when I make a change or that it doesn't work as the deem it should etc. If I break too many things, people start handing out bad advice about not updating or installing some dodgy fork/repo/patch which overrides my version.

Even if I shrug all that off, it still impacts me. I end up wasting time fielding questions and looking into bug reports only to find out the issue is related to some weird gist patch the user installed from a random Reddit user to "make it work" with another addon.

MoojMidge:
> Even if I shrug all that off, it still impacts me

Yeah, people do tend to have a massive sense of entitlement. I think it gets worse when they don't understand the time/effort/difficulty involved in doing something (not dissimilar to the problems in the retail/hospitality/service industry really, but at least people get paid there). Burnout can happen to anyone, I can only hope that it doesn't get that far.

As for this PR, aside from not really listening to you in the first place about how you envisaged making only certain modules available, I realised I was going about this in entirely the wrong way, but I think I have now addressed your concerns.

Rather than dynamically importing a bunch of modules into the module extension endpoint namespace, to make them available for importing elsewhere, what I should have done is use a [meta path import hook](https://docs.python.org/3/reference/import.html#import-hooks) and just leave everything else up to the normal Python import system.

This way, there is no hard coded imports into the module extension endpoint (and no dynamic imports either), no need to create wrapper classes that are imported even if they are not used, or modify existing code (apart from what has already been done to separate the API keys).

Control of what is available to be imported is done via the [permissions module](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/pull/983/files#diff-97685b70e9746a75598efc0b2ad84a0d6c6b3e29be846506f40afb44d8c3f44eR6-R40
) which allows for fine grained control:
- only allowed modules can be imported
- but not if they are denied
- an allowed module can be restricted to require an enabled user setting to prevent api keys/token from being imported even if the module is imported

Let me know if this is a more palatable solution, otherwise I can close this PR and create a new one with only the hardcoded imports and the wrapper classes.

jurialmunkey:
Okay this looks really promising based on a quick look over. I've hit a bit of a busy patch at work, so might take me a week or so to review properly but this is looking really good (and nice and clean too). Really appreciate all the effort that has gone in here.

jurialmunkey:
Side note for later: Realised I need to convert my settings.xml from tabs to spaces -- didn't notice until this PR that it's using tabs. Likely crept in when I used Ronie's script to convert from the old settings format.

jurialmunkey:
I should be able to have a proper look at this over the weekend. I see you're still adding some commits. Do you want me to wait to review?

MoojMidge:
Last commit was a bug that I didn't notice previously, the rest are all from previous changes, just rebasing and force pushing. Don't anticipate any more, pending your review. No rush. 

Let me know if you want me to fix the whitespace in settings.xml

resources/language/resource.language.en_gb/strings.po
```diff
@@ -2650,7 +2650,7 @@ msgstr ""
 
 #: /resources/lib/requestapi.py
 msgctxt "#32008"
-msgid "API Key"
+msgid "Allow 3rd party plugin access to API key/token"
```
jurialmunkey: I know this ID is currently unused but you're better to generate a new ID. This ID already has translations, so if you're going to change it then you also need to change all the translations too.

resources/settings.xml
```diff
@@ -409,6 +409,11 @@
 						<heading>32001</heading>
 					</control>
 				</setting>
+                <setting id="omdb_apikey_access" type="boolean" label="32008" help="" parent="omdb_apikey">
```
jurialmunkey: See comment above about localization. Same goes for the other settings with this label ID.

resources/settings.xml
```diff
@@ -421,6 +426,11 @@
 						<heading>32423</heading>
 					</control>
 				</setting>
+                <setting id="mdblist_apikey_access" type="boolean" label="32008" help="" parent="mdblist_apikey">
```
jurialmunkey: Same

resources/settings.xml
```diff
@@ -448,6 +458,11 @@
 					<default>false</default>
 					<control type="toggle"/>
 				</setting>
+                <setting id="fanarttv_clientkey_access" type="boolean" label="32008" help="" parent="fanarttv_clientkey">
```
jurialmunkey: Same

resources/settings.xml
```diff
@@ -502,6 +507,21 @@
...
+					<level>0</level>
+					<data>RunScript(plugin.video.themoviedb.helper,authenticate_trakt)</data>
+					<constraints>
+						<allowempty>true</allowempty>
+					</constraints>
+					<control type="button" format="action">
+						<close>true</close>
+					</control>
+				</setting>
+                <setting id="trakt_token_access" type="boolean" label="32008" help="" parent="authenticate_trakt">
```
jurialmunkey: Same

jurialmunkey (changes requested):
Some very minor changes in comments about localization ID but otherwise I think this looks great. Code is super clean and readily understandable.

I'm going to try out a few test cases to make sure nothing is broken (doubtful) and works as intended (expected) but I'm otherwise pretty happy to merge once the localization is fixed.

jurialmunkey:
Hmmm. I'm getting an error when I try to import from another addon:

```
File "C:\Users\coope\AppData\Roaming\Kodi\addons\script.skinvariables\resources\lib\method.py", line 15, in do_test
 from themoviedb_helper.api.tmdb.api import TMDb
File "C:\Users\coope\AppData\Roaming\Kodi\addons\plugin.video.themoviedb.helper\resources\modules\themoviedb_helper\__init__.py", line 46, in <module>
 from resources.lib.addon.permissions import __access__
ModuleNotFoundError: No module named 'resources.lib.addon'
-->End of Python script error report<--
```

~Base path of module dir is different from resources.lib~
~Appears `self._Addon(addon_id).getAddonInfo('path')` is returning empty~
EDIT Scrap that. Forgot to change log level when I was logging the output of getAddonInfo(). 

Not sure what the issue is here.

jurialmunkey:
Okay I think it is because the basedir is not initialised as a module and Kodi prevents it from being used like that (at least on Windows on Py3.8.5). So if we shift base dir to resources *then* we can import

See this branch: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/compare/MoojMidge-nexus_module_alternative_approach

This commit https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/86aafb869819f8b10cb724cbaaffa1bcdb4d8eaf shifts from `resources.lib` to just `lib` and appears to work -- but will need more extensive testing first to make sure I didn't break anything. I simply did a bulk find and replace of `resource.lib` and moved base scripts up to resources.

Main *real* change I made is to the `__init__.py` in the module itself. That was simply to change basedir to resources:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/86aafb869819f8b10cb724cbaaffa1bcdb4d8eaf#diff-87c3967f8446db14e4a2a544ec50db34b9d4a7843e2c43500757439d46c34fbc

That worked for me at least for first test of importing `TMDb()`

EDIT: After further testing, everything appears to be working great for me now with these changes. 

- Imports are being allowed/denied appropriately. 
- API Keys are not being imported unless permitted for base api module -- e.g. TMDb() needs key
- Imported modules can be used as intended with internal API key -- e.g. Players() module will still be usable

MoojMidge:
Hmm that is strange. It shouldn't be necessary to make that change, as there should be no difference whether the path is being searched from `...\plugin.video.themoviedb.helper\resources\` or `...\plugin.video.themoviedb.helper\` - they are both namespace packages and should be searched in the same way.

I think what is happening is that this PR creates a meta path import hook to check for the right access permissions, but still uses the standard path based finder to actually find modules. The standard path based finder uses the current working directory to search for a given path.

When I was testing this I was using UpNext, and the current working directory for the script and service entry points was `...\service.upnext\resources\lib`. When the path based finder did a search for the `\resources\lib\` path it did not find anything in the current working directory i.e. there was no `...\service.upnext\resources\lib\resources\lib\` path found, and so `sys.path` was checked allowing the `...\plugin.video.themoviedb.helper\` entry to be used.

When you are testing using `script.skinvariables` the current working directory would be `...\script.skinvariables\`, and this parent path does have a `\resources\lib` directory, which the path entry finder takes as an indication that this is the correct path to use to search for the required module.

Changing the path for the script and service entry points to  `...\plugin.video.themoviedb.helper\resources\` masks the problem when importing from `script.skinvariables` in the same way that it was masked when I was originally testing i.e. there is no `...\script.skinvariables\lib\` directory so `sys.path` gets checked instead. However the problem will re-occur if the import was to occur from another addon where the current working directory does have a `\lib\` sub-directory.

I think I will need to do some reading about how to implement a path entry hook.

MoojMidge:
Thanks for testing btw, I think the internal import and permission/access side of things works properly, just need to figure out the correct way to handle the external import.

I might set this back to draft in the meantime, until I get some time to look at python docs.

jurialmunkey:
> Changing the path for the script and service entry points to ...\plugin.video.themoviedb.helper\resources\ masks the problem when importing from script.skinvariables in the same way that it was masked when I was originally testing 

This makes a lot of sense. Tested and can confirm moving skinvariables from `resource/lib` to `lib` will cause the issue again.

To complicate matters there's also this `__init__.py` trap: http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-init-py-trap

SkinVariables has `__init__.py` in `resources/lib` which is what blocks the original import in `themoviedb_helper/__init__.py`

if I remove the `__init__.py` from `skinvariables/resources/lib` it manages to properly import the module I want but then fails when that module needs something in that namespace.
```
File "C:\Users\coope\AppData\Roaming\Kodi\addons\script.skinvariables\resources\lib\method.py", line 12, in do_test
 from themoviedb_helper.player.players import Players
File "C:\Users\coope\AppData\Roaming\Kodi\addons\plugin.video.themoviedb.helper\resources\lib\player\players.py", line 11, in <module>
 from resources.lib.items.listitem import ListItem
ModuleNotFoundError: No module named 'resources.lib.items'
```

So I think there's two separate issues in play here. The `__init__.py` trap and the shared namespace naming.

jurialmunkey:
@MoojMidge - Simple solution. Move `resources/lib` to `resources/tmdbhelper/lib` and then shift base path up to `plugin.video.themoviedb.helper/resources` -- now we have a unique module name in the modules namespace: `tmdbhelper.lib`

Commit: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/e2e4a583ef34dc0dab8802cbae6dd093b4a9ae2a
Branch: https://github.com/jurialmunkey/plugin.video.themoviedb.helper/tree/nexus_module_alt

This appears to work well for me.

Additionally, I was thinking since it's adding the module to path, then it could go anywhere. Perhaps it would be better for the module `__init__.py` code go into `script.module.tmdbhelper` https://github.com/jurialmunkey/script.module.tmdbhelper

Then it is one import. If the user wants to get basic functions from the module they use `tmdbhelper.` prefix and if they want libs from plugin.video.themoviedb.helper then they use `tmdbhelper.lib.` prefix

MoojMidge:
> Simple solution

Avoiding unnecessary complexity is a good idea, so if you are willing to change the addon structure then that should work well.

> Additionally, I was thinking since it's adding the module to path

I was naively thinking that I could avoid path/namespace pollution by adding and removing from the path, but didn't consider that there may be an existing path that could result in a collision...

> Perhaps it would be better for the module `__init__.py` code go into `script.module.tmdbhelper` https://github.com/jurialmunkey/script.module.tmdbhelper
> 
> Then it is one import

~~Good idea. The only thing that may need to be considered is the need for additional error handling to check that the `xbmc.Addon()` call returns correctly (`plugin.video.themoviedb.helper` is installed and is the correct version. Currently I do this implicitly using the `version` attribute of the `import` tag in my `addon.xml`, but moving it to `script.module.tmdbhelper` would mean it should be done in `__init__.py` itself) and avoid adding the meta path import hook if it doesn't.~~

~~Perhaps in the `SysPathLoader.__init__()` can check that the renamed base library directory exists, indicating whether the user has the correct version of `plugin.video.themoviedb.helper` installed, and handle things appropriately from there.~~

Unfortunately this is not going to work.

Because the `tmdbhelper.lib.` prefix is used internally in `plugin.video.themoviedb.helper`, it means that all the internal imports for the script/plugin/service entry points now get processed by `script.module.tmdbhelper`. This can be worked around by adjusting the permissions, but it is additional overhead for every import.

The bigger problem is that the content of `sys.path` is different when importing from `script.module.tmdbhelper`, for example `script.module.infotagger` is a dependency of `plugin.video.themoviedb.helper`, but not for `script.module.tmdbhelper`, so an import like `from infotagger.listitem import ListItemInfoTag` will fail when it was initiated through `script.module.tmdbhelper`. This could probably be worked around as well, but would add more overheads to every import.

resources/language/resource.language.en_gb/strings.po
MoojMidge: New 32483 string ID used, 32008 reverted to original content.

jurialmunkey:
Realised the same about the circular import and name collision of tmdbhelper.lib yesterday - you beat me to mentioning it won't work.

Including in the existing module was really only a passing idea of something that might've been easy to add as a nice feature. Definitely not worth complicating the approach over it.

It looks like you've got a few file conflicts in the most recent changes. I think a rebase against nexus branch was needed before restructuring (I forgot to do that myself).

MoojMidge:
> It looks like you've got a few file conflicts in the most recent changes. I think a rebase against nexus branch was needed before restructuring (I forgot to do that myself).

Yeah local repo was out of date. All good now

jurialmunkey:
Okay I've merged because fairly sure underlying code works well and I'm big on "Release Early, Release Often" approach to squashing bugs. Also don't want to give us the headache of constantly rebasing.

Thanks so much for sticking with this PR. The interesting question will be whether I meet resistance on the official repo about modifying sys.path -- official repo updates are pretty slow at the moment due to a gap in the team, so that's a problem for another day anyway.

MoojMidge:
Maybe the system.path manipulation won't be an issue because it only happens if the module extension point is executed, and the additional path is removed after it is used. Let's wait and see I guess.

One thing about the v5.1.0 release notes:

> Currently this PR opens up:
> - from tmdbhelper_lib.player.players import Players
> - from tmdbhelper_lib.api.tmdb.api import TMDb
> - from tmdbhelper_lib.player.details import get_next_episodes

The entire player package is actually made available with the final iteration of the PR.

Also noticed a question about the new settings. Should I make them hidden for the moment as they are not used? 

Thanks again for the input and testing and making this available in the first place.

jurialmunkey:
> Maybe the system.path manipulation won't be an issue because it only happens if the module extension point is executed, and the additional path is removed after it is used.

Yeah thats what I'm banking on. It's a little unorthodox but I think the approach is reasonable and with the restructure I can't really see how it would cause any issues. 

> 
> One thing about the v5.1.0 release notes:
> 
> > Currently this PR opens up:
> > 
> > * from tmdbhelper_lib.player.players import Players
> > * from tmdbhelper_lib.api.tmdb.api import TMDb
> > * from tmdbhelper_lib.player.details import get_next_episodes
> 
> The entire player package is actually made available with the final iteration of the PR.

Ah I meant to check that. I think I just quickly grabbed the paths from one of the commit messages. I'll update the release notes. Really need to make a Wiki page documenting the features. 


> Also noticed a question about the new settings. Should I make them hidden for the moment as they are not used?

Yeah I think might be good to hide for the moment to avoid unnecessary questions. 

If youre going to PR that change, worthwhile to convert the tabs to spaces at same time.

MoojMidge:
All good, I'll create a PR over the weekend

PR #1011, opened 2023-02-19
Don't call getUniqueID if function not defined

Kayzels:
Older versions of Kodi (Matrix and older) do not have a defined function
called getUniqueID for video tags.

This checks whether the function exists, and otherwise sets the tmdb_id
to None, which will then revert to the old behaviour later.

This fixes the issue in Arctic Horizon 2, where clearlogo isn't populating during playback. It is a bit of a hacky fix, though.

### DIFF
--- a/resources/lib/monitor/player.py
+++ b/resources/lib/monitor/player.py
@@ -69,8 +69,11 @@ class PlayerMonitor(Player, CommonMonitorFunctions):
         self.epyear = info_tag.getYear() if self.dbtype == 'episode' else None
         self.season = info_tag.getSeason() if self.dbtype == 'episode' else None
         self.episode = info_tag.getEpisode() if self.dbtype == 'episode' else None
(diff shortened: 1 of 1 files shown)


jurialmunkey:
Ah! I didn't spot that it was v20 addition.

No need for the hasattr check as there's a separate matrix branch which can have its own changes https://github.com/jurialmunkey/plugin.video.themoviedb.helper/tree/matrix

Do you know if the infolabel works in Matrix? Because according to skinning manual it does (which is why I thought it'd be in info tag)
`$INFO[VideoPlayer.UniqueID(tmdb)]`

If the infolabel works I can retrieve it with get_infolabel()

Kayzels:
From what I can tell, VideoPlayer.UniqueID exists in Matrix, but it doesn't take in a parameter.

So UniqueID(tmdb) doesn't seem to work. The video player is not displaying the clearlogo.

Do you want me to stay on Matrix, so you have a tester with the environment? I've been meaning to update.

PR #1018, opened 2023-03-05
Honor the order of requested SourceImage lookups

pakraider:
An infolabel shouldn't override any artwork lookups that are listed before it. For example, Art(fanart)|EPGEventIcon should attempt to lookup Art(fanart) before falling back to EPGEventIcon.

### DIFF
--- a/resources/tmdbhelper/lib/monitor/itemdetails.py
+++ b/resources/tmdbhelper/lib/monitor/itemdetails.py
@@ -94,23 +94,23 @@ class ListItemDetails():
     def get_artwork(self, source='', build_fallback=False, built_artwork=None):
         source = source.lower()
         infolabels = ARTWORK_LOOKUP_TABLE.get(source, source.split("|") if source else ARTWORK_LOOKUP_TABLE.get('thumb'))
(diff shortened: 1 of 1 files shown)


jurialmunkey:
Thanks for taking the time to make this PR but unfortunately I disagree with this change. Local artwork should always be preferred over online artwork.

For instance, with the fallback order: `Art(clearlogo)|Art(tvshow.clearlogo)` some episodes or seasons might have local art set `clearlogo` but others might only have local art for `tvshow.clearlogo`. Generally users will want their local tvshow.clearlogo to have precedence over an online clearlogo.

TMDbHelper is used in several skins, so such a significant change to core behaviour will impact far more than AH2. If you want AH2 to have an option not to use EPGEventIcon then I'm happy to accomodate that request in the skin.

pakraider:
That's a good point - the blast radius for this would be pretty vast. Thanks for the explanation!

I was actually using this for AH2 to have EPGEventIcon as a fallback for FlixArt in the EPG, but to prioritize online lookups. I'll open a request over there. :)

PR #1057, opened 2023-06-05
Option to separete Trakt lists into folders

tielis:
Hi Jurialmunkey,

First of all, thank you so much for all the amazing work you have done for the Kodi community. 

I am submitting this pull request because I am facing the following issue. I have three Trakt Lists arranged into "movies & tv shows", "animes & cartoons", and "documentaries", of which each of them is pretty long, with more than 1000 items each.

But,  I've found that the current way TMDB Helper is managing Trakt Lists through creating a smart playlist for each list and setting a rule individually for each str. file, at least here in my Nvidia Shield, it takes quite a long time every time I open the playlist (Node editor).

However, when we have the Lists separated into folders, the smart playlist rule, path based, is much easier for Kodi to handle. Therefore, when we try to open it, the list opens instantly.

I am not a programmer,  and I cannot understand most of your code, LoL. So I am not 100% sure of what I wrote. Feel free to rewrite whatever you want. You can also ask me to fix it, no problem at all.

Even if you don't agree with this PR, I kindly ask you to consider implementing this feature (trakt list into folders), this would be, IMO, a huge improvement for the users who uses trakt lists  and TMDB Helper to manage Kodi library. 

PS: I used the v5.1.8 because I couldn't find the file "jurialmunkey" for the logger in the v5.1.15 

Thank you

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ target/
 .idea/
 
 *.zip
+/.vs/plugin.video.themoviedb.helper/FileContentIndex
+/.vs/plugin.video.themoviedb.helper/v17
+/.vs
--- a/addon.xml
+++ b/addon.xml
@@ -10,7 +10,7 @@ provider-name="jurialmunkey">
   <import addon="script.module.addon.signals" version="0.0.6" />
   <import addon="script.module.tmdbhelper" version="0.0.8" />
   <import addon="script.module.infotagger" version="0.0.5" />
-  <import addon="script.module.beautifulsoup4" version="4.9.3" />
+  <import addon="script.module.beautifulsoup4" version="4.9.3" />  
 </requires>
 <extension point="xbmc.python.pluginsource" library="resources/plugin.py">
   <provides>video</provides>
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -2782,3 +2782,8 @@ msgstr ""
 msgctxt "#32027"
 msgid "Clear default players"
 msgstr ""
+
+#: /resources/settings.xml
+msgctxt "#32490"
+msgid "Separate Trakt list(s) into folder(s)"
+msgstr ""
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -334,6 +334,16 @@
                         <data>RunScript(plugin.video.themoviedb.helper,monitor_userlist)</data>
                     </control>
                 </setting>
+                <setting id="trakt_list_folders" type="boolean" label="32490" help="" parent="library_autoupdate">
+                    <level>0</level>
+                    <default>false</default>
+                    <dependencies>
+                        <dependency type="enable">
+                            <condition operator="is" setting="library_autoupdate">True</condition>
+                        </dependency>
+                    </dependencies>                    
+                    <control type="toggle"/>                    
+                </setting>
                 <setting id="monitor_userslug" type="string" help="">
                     <level>0</level>
                     <default/>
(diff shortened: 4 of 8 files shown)


jurialmunkey:
Splitting lists into separate folders is a bad idea. What happens if two of your lists have the same item? Either you're duplicating items in your library, or one of your smart playlists is missing an item.

I'm not sure I understand how having a smart playlist with 1000s of items in it is useful versus just using normal library navigation? How does that help you find something to watch? If all you're doing is splitting by documentary and animation, then you can already browse those via library genres without needing a smart playlist.

The smart playlist functionality is more intended for specific lists like watchlist or special interests (e.g. I have some lists like 80s action; sci-fi classics; studio ghibli etc.). It's not really intended to add every single movie in existence.

tielis:
Yes, I agree that duplicating items is a big issue. But this will all depend on how the user uses trakt lists. That is why I set it as an optional feature and "false" by default. 

If somehow this would help convince you, I could try to set a popup message whenever the feature is toggled on, warning about duplication. (I am not sure if I know how to do that, Lol.)

In the end, all I am trying to do is to give the users more options to manage the library.

Using gender to build the playlists was my first attempt to organise my library, but that didn't work quite well.
Mainly because, at the moment, Kodi and TMDB don't have an "Anime" or "cartoon" gender. Both have only "animation", which includes 3D movies, and imo they are completely different species. Also, my movies and Tv shows were always a mess mixed with all the animes and cartoons I have.

To find something to watch, I have my "animes & cartoons" path excluded from the Gender node, I used the "Library Node Editor" to do that. This way, when I want to find an action movie to watch, I don't have, for example, "Castle in the Sky" or " "The Lion King II" popping up. On the other hand, when I want to watch an anime, I don't really need to use gender to pick one. But I could easily create a node too.

I know this can all be done with playlists, but it takes much longer to load them. However, when all is set as path-based, everything loads instantly.

Organising the library through the Trakt website is much easier, as it has a more powerful filter with many options and a better UI. Also, It is all stored in the cloud, so you only need to do it once.

Anyway... hopefully, I've managed to change your mind, Lol. Otherwise, I will have to keep using my fork and rebasing every now and then :(

Again, thanks a lot Jurialmunkey for your great work.

jurialmunkey:
Being optional or having warnings doesn't address the underlying issue. Splitting the library into folders based on metadata isn't a good approach to file system categorisation imho. For instance, what happens if something is both animation and a documentary?

There's a lot of different ways the library can be split up virtually using metadata without needing to separate the file location physically. For instance, a separate routine could run to add a tag with the list name, then you just browse by tag.

The only reason the smart playlist currently is slow is because it adds each individual file path, and it was only done that way because the intention was that users would make smaller more curated lists rather than trying to bulk add everything in one massive list.

jurialmunkey:
@tielis - Please try v5.1.16

After a library update, TMDbHelper will go through and tag every item in a monitored list with tags matching the username and list slug. 

That should give you a series of library tags like:
"Trakt User jurialmunkey"
"Trakt List watchlist"
etc.

You can then go directly into the tags section of the library to access these. I also update the smart playlists to use the tags instead so you don't need to worry about updating shortcuts.

Please let me know if it works well for your large library. I expect this approach should display items much faster because the smart playlists will only have two rules (tag = username and tag = listname).

tielis:
Omg, thank you much jurialmunkey. You are a legend.

I'm optimistic this update will perform much faster. And if everything works fine, using tags is a much better approach, I agree.

I I'll play with it on the weekend and let you know it goes.

Thnaks

PR #1127, opened 2023-10-28
:bug: Fix recommendations onback history bug

### LINKED ISSUE #1120: [BUG] Back to recommendations dialog
### Describe the bug

When I enter the recommendation window, then open the info about the actor or some other info about the movie and then return to the basic dialog info, retrying to enter the recommendation window is no longer possible. It is interesting that this happens with the android device but not on the windows platform. The same thing happens with the AH2 skin.

### Steps To Reproduce

//

### Debug log

https://paste.kodi.tv/inubonezod

//

### Checklist

- [X] I have searched the issues and this bug has not been reported.
- [X] I have updated TMDbHelper to the most recent version.
- [X] I have described the bug and included steps to reproduce the issue
- [X] I have uploaded my debug log.

jurialmunkey:
Possible candidate for resolving onback bug in recommendations.

Resolves #1120

### DIFF
--- a/resources/tmdbhelper/lib/window/recommendations.py
+++ b/resources/tmdbhelper/lib/window/recommendations.py
@@ -322,10 +322,10 @@ class WindowRecommendationsManager():
             data = self._current_dump
             self.open_recommendations()
 
+            # If RECS closed because ONBACK then we need to go back to the previous info and join it
             if self._gui._state == 'onback':
                 self._gui = _gui
                 self._current_path = path
-                self._current_dump = get_property(PROP_JSONDUMP, set_property=data)  # CHECK HERE
                 return self.on_join(t, path)
 
         return self.on_back() if self._history and not self.is_exiting() else self.on_exit()
@@ -384,6 +384,8 @@ class WindowRecommendationsManager():
         if xbmcgui.getCurrentWindowId() != self._window_id:
             executebuiltin(f'ActivateWindow({self._window_id})')
             self.wait_until_active(self._window_id, poll=0.1)
+        self._current_dump = ''
+        get_property(PROP_JSONDUMP, clear_property=True)
         if threaded:
             t = Thread(target=xbmcgui.Dialog().info, args=[listitem])
             t.start()


jurialmunkey:
@matke-84 @roidy 

I believe this PR will fix the onback issue in recommendations. However, it requires testing before merging to confirm that it does not cause other issues.

If you could please test and let me know if it does cause any **new** issues that don't currently exist.

roidy:
@jurialmunkey Thanks, but unfortunately this doesn't work, opening any item from the recommendations window results in a blank info dialog.

https://streamable.com/l4wu5w 

The default nexus branch also shows this issue.

jurialmunkey:
@roidy - Note that latest master branch has a dependency bump for script.module.jurialmunkey -- you need v0.1.10 of the module installed
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/1503ccd4c1a706be7447db3ab86754da42d19731/addon.xml#L11

https://github.com/jurialmunkey/script.module.jurialmunkey

roidy:
@jurialmunkey Sorry, my bad.

I can confirm that after a few minutes of limited testing it does indeed fix the issue of not being able to return to the recommendations window from the initial info dialog 👍 I can't see any new issues but I've only tested for a few minutes, I'll get a chance to do more testing later on today.

matke-84:
@jurialmunkey I tested it quickly now. I would say that everything works as it should. I see that now the recommendation window closes also quickly. What can I say but thank you. Thank you for this fantastic addon and everything you do for us skinners. I couldn't imagine my skin without TMDB Helper. Thanks again.
P.S. If I notice some bug later I will report it.

jurialmunkey:
Thanks for testing guys. 

I think I will merge now, as I'm fairly confident it is handling things correctly.

The code itself is not that complex. It's just that combining native info dialog with scripted recommendations makes things a bit chess-like thinking through all the possible permutations of "moves" a user might make and how to handle each scenario -- it'd be much simpler to handle that if the info dialog were scripted like extendedinfo but that has its own limitations.

matke-84:
Unfortunately, the problem has returned. I don't know with which version, I haven't used kodi for a while. Tested with ah2 skin:

https://streamable.com/w8dwaa

roidy:
@matke-84 Still working fine for me with:-

plugin.video.themoviedb.helper - 5.2.15
script.module.jurialmunkey - 0.1.17

matke-84:
@roidy Can you please try to do the same as I did in the video? The same thing happens in ah2 and my skin. Latest version of tmdb helper.

jurialmunkey:
Issue has always existed. I have not touched the recommendations dialog management code since this PR, and this issue is different from what this PR fixes (superficially similar; very different cause).

If you close the info dialog and reopen it, it will resolve itself.

I'm not sure if I'm willing to go down this rabbit-hole again. I'm not sure if a "perfect" experience with scripted recommendations window is possible without core changes to how Kodi manages dialogs. 

It will always be a hack workaround with a few rough edges.

PR #1175, opened 2024-03-14
Rewrite msgctxt #32481

dm1tz:
Also, update strings.po uk_ua.

### DIFF
--- a/resources/language/resource.language.ar_sa/strings.po
+++ b/resources/language/resource.language.ar_sa/strings.po
@@ -2528,9 +2528,7 @@ msgstr ""
 
 #: resources/settings.xml
 msgctxt "#32481"
-msgid ""
-"Increases the pool of recently watched items randomly selected from to use "
-"as the seed for becauase you watched style recommendations."
+msgid "Increases the pool of recently watched items randomly selected from the seed to use for \"Because You Watched\" style recommendations."
 msgstr ""
 
 #: resources/settings.xml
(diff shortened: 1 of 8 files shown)


dm1tz:
The original version seems a bit elusive to me. Hope that the edit makes sense.

jurialmunkey:
Yep change makes sense. Just need to escape the quotes around "Because you watched" in the msgid.

dm1tz:
> Yep change makes sense. Just need to escape the quotes around "Because you watched" in the msgid.

Escaping the quotes is optional I believe. The [gettext manual](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) states:
```
The two strings, untranslated and translated, are quoted in various ways in the PO file, 
using " delimiters and \ escapes, but the translator does not really have to pay attention 
to the precise quoting format, as PO mode fully takes care of quoting for her.
```
Also, I haven't noticed any errors or warnings in the Kodi log. But still can do it if you want.

jurialmunkey:
That's talking about breaking the string over multiple lines which doesn't matter how you do it as po just joins it all back together.

But if you want to use a double quote *inside* the string then you should escape it to make sure it displays correctly and is not read as an enclosing string character.

dm1tz:
> That's talking about breaking the string over multiple lines which doesn't matter how you do it as po just joins it all back together.
> 
> But if you want to use a double quote _inside_ the string then you should escape it to make sure it displays correctly and is not read as an enclosing string character.

Thanks for the clarifications.

PR #1180, opened 2024-03-18
Small update for artwork quality

matke-84:
For landscape and thumbs, I don't think the whole screen will ever be used, usually skins use tiles for these two artworks.
I put the original for the poster because on higher resolution screens, whoever chooses original artwork quality will get a better image quality.


resources/tmdbhelper/lib/addon/consts.py
```diff
@@ -38,9 +38,9 @@
 IMAGEPATH_SMALLLOGO = 'https://image.tmdb.org/t/p/w300'
 IMAGEPATH_NEGATE = 'https://image.tmdb.org/t/p/h100_filter(negate,000,666)'
 IMAGEPATH_ALL = [IMAGEPATH_ORIGINAL, IMAGEPATH_HIGH, IMAGEPATH_LOW, IMAGEPATH_POSTER, IMAGEPATH_SMALLPOSTER, IMAGEPATH_SMALLLOGO]
-IMAGEPATH_QUALITY_POSTER = [IMAGEPATH_POSTER, IMAGEPATH_POSTER, IMAGEPATH_POSTER, IMAGEPATH_SMALLPOSTER]
+IMAGEPATH_QUALITY_POSTER = [IMAGEPATH_ORIGINAL, IMAGEPATH_POSTER, IMAGEPATH_POSTER, IMAGEPATH_SMALLPOSTER]
```
jurialmunkey: There is no reason to get original sized posters as Kodi will simply downscale anyway with default settings
https://kodi.wiki/view/Advancedsettings.xml#imageres

At default, artwork max height for poster aspect is 720p. Anything taller is downscaled by Kodi before caching.
At w500, the IMAGEPATH_POSTER value will limit posters 500x750 (2:3 standard ratio) which is just about as close as possible to Kodi's default setting.

resources/tmdbhelper/lib/addon/consts.py
```diff
@@ -38,9 +38,9 @@
...
 IMAGEPATH_QUALITY_FANART = [IMAGEPATH_ORIGINAL, IMAGEPATH_HIGH, IMAGEPATH_HIGH, IMAGEPATH_LOW]
-IMAGEPATH_QUALITY_THUMBS = [IMAGEPATH_ORIGINAL, IMAGEPATH_HIGH, IMAGEPATH_HIGH, IMAGEPATH_LOW]
+IMAGEPATH_QUALITY_THUMBS = [IMAGEPATH_ORIGINAL, IMAGEPATH_HIGH, IMAGEPATH_LOW, IMAGEPATH_POSTER]
```
jurialmunkey: w500 (IMAGEPATH_POSTER) is not valid for backdrop artwork
https://www.themoviedb.org/talk/53c11d4ec3a3684cf4006400

Only original, w1280 (high), w780 (low) and w300 are valid. And w300 is so ridiculously tiny to not be worthwhile imo.

jurialmunkey: > For landscape and thumbs, I don't think the whole screen will ever be used, usually skins use tiles for these two artworks.

I use episode thumbs for episode backgrounds.

matke-84: Theoretically and according to the table it is not, but practically it is.
https://image.tmdb.org/t/p/w500/1e1mvQmgNAelU1xhgi9EoEv8u9z.jpg

resources/tmdbhelper/lib/addon/consts.py
matke-84: Ok.

resources/tmdbhelper/lib/addon/consts.py
matke-84: I went to match it with the other artwork quality. Also we would gain a little in speed, and when a large image is downscaled in Kodi, it usually looks a little worse.
Thumbs are usually of lower quality, so they are rarely used for the background. But ok.

jurialmunkey: > Theoretically and according to the table it is not, but practically it is. https://image.tmdb.org/t/p/w500/1e1mvQmgNAelU1xhgi9EoEv8u9z.jpg

I'm hesitant to use undocumented parts of the API because Travis Bell at TMDB has made it clear in the past that undocumented features could be removed at any time and shouldn't be relied upon.

>  we would gain a little in speed

I'd say probably unlikely to be noticeable. The difference in filesize from w780 to w500 is only 29kb. 

image

At smaller resolutions there is less gain because change in surface area is less dramatic --
780x438 to 500x281 only reduces by ~200 thousand pixels
4k (3840x2080) to 1280x720 reduces by ~7 million pixels.

If anything, the w500 size is likely to take longer than w780 because it is not a standard size and so less likely to be in the TMDb server cache and the resized image will need to be generated first.

> when a large image is downscaled in Kodi, it usually looks a little worse.

I definitely agree (at least with default settings) but I think w780 is probably not too bad here.

Remember w780 is width, not height. So at 16:9 it is only 438 high, which is only 40% of height of a 1080p screen (and 20% of 4K).

matke-84: > Remember w780 is width, not height. So at 16:9 it is only 438 high, which is only 40% of height of a 1080p screen (and 20% of 4K).

For medium, I think it's completely ok. :-)
And please just check this out when you have time. https://github.com/jurialmunkey/plugin.video.themoviedb.helper/issues/1181

PR #1187, opened 2024-03-28
Add focus_id to recommendations window

roidy:
Add ability to set a custom control to be focused:-
                    `Runscript(plugin.video.themoviedb.helper,recommendations=5000|info=cast|true|info,window_id=1191,focus_id=4999,......)`

### DIFF
--- a/resources/tmdbhelper/lib/window/recommendations.py
+++ b/resources/tmdbhelper/lib/window/recommendations.py
@@ -73,6 +73,7 @@ class WindowRecommendations(xbmcgui.WindowXMLDialog):
         self._queue = (i for i in self._recommendations)
         self._context_action = kwargs.get('context')
         self._window_id = kwargs['window_id']
+        self._focus_id = kwargs.get('focus_id')
         self._window_manager = kwargs['window_manager']
         self._window_properties = {
             k.replace('winprop_', ''): v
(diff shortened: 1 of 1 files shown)


jurialmunkey:
This is a reasonable change. Can be better placed in the code, however.

The `list_id` arg in the `_build_all_in_groups` func *is* the argument telling it which ID to focus, so you shouldn't need to override its behaviour with an if clause. Instead, simply pass it a different ID when the func gets called.

It gets called here:

https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/ed1e2d2e9c3d253e3130c890f5ce9c1aa542d9d9/resources/tmdbhelper/lib/window/recommendations.py#L98-L100

So simply add a line inbetween (line 99) to override the list_id that gets passed with the focus_id like so:
```python
 _list_id = self._add_items(_next_id, _listitems) 
 _list_id = self._focus_id or _list_id  # Allow skinner to override first list default focus
 Thread(target=self._build_all_in_groups, args=[3, _list_id]).start()  # Don't block closing 
```

roidy:
I see, I'm not familiar with the code so it was a quick hack.  

However couldn't it be optimised further, we shouldn't need to assign `_list_id`  twice:-

```
 _list_id = self._focus_id or self._add_items(_next_id, _listitems) # Allow skinner to override first list default focus
 Thread(target=self._build_all_in_groups, args=[3, _list_id]).start()  # Don't block closing 
```

jurialmunkey:
> I see, I'm not familiar with the code so it was a quick hack.
> 
> However couldn't it be optimised further, we shouldn't need to assign `_list_id` twice:-
> 
> ```
>  _list_id = self._focus_id or self._add_items(_next_id, _listitems) # Allow skinner to override first list default focus
>  Thread(target=self._build_all_in_groups, args=[3, _list_id]).start()  # Don't block closing 
> ```

Then the add_items func will not be called if focus_id has a value.

roidy:
Fixed.

jurialmunkey:
lgtm

PR #1230, opened 2024-07-27
:zap: Allow overriding close action in recs window for specific lists

### LINKED ISSUE #1229: [Feature] Recommendations override onBack
### Describe the feature that you'd like

Could we get the ability to override `onback` for containers in the recommendations dialog like the old embuary info script. https://github.com/sualfred/script.embuary.info?tab=readme-ov-file#overwriting-onback-or-provide-an-additional-onclose-action

### Why is this feature needed? What problem does it solve?

Sometimes you don't want to close the recommendations dialog when pressing `onback` on a list. I would like to use `onback` to reset a list to the first item and only close the recommendations dialog when the user scroll up past the top of the group of lists.

### Steps to reproduce.

_No response_

_No response_

### Checklist

- [X] I have searched the issues and this feature has not been requested.
- [X] I have updated TMDbHelper to the most recent version.

jurialmunkey:
Resolves #1229

Adds additional "close" argument to recommendations args:
```
recommendations=list_id(int)|paramstring(str)|related(bool)|action(str)|close(str)
```

Will override the close action with the builtin specified for that list. Additionally, can use window properties in same way as `action(str)` for undefined lists e.g.

```
<onload>SetProperty(Close_{list_id},{builtin})</onload>
```

### DIFF
--- a/resources/tmdbhelper/lib/window/recommendations.py
+++ b/resources/tmdbhelper/lib/window/recommendations.py
@@ -31,12 +31,13 @@ ID_VIDEOINFO = 12003
 
 """
 Runscript(plugin.video.themoviedb.helper,recommendations=)
-recommendations=list_id(int)|paramstring(str)|related(bool)|action(str) [Separate multiples with || ]
+recommendations=list_id(int)|paramstring(str)|related(bool)|action(str)|close(str) [Separate multiples with || ]
     * The lists to add. Separate additional lists with ||
     * list_id: the container that the items will be added
     * paramstring: the tmdbhelper base path such as info=cast
     * related: whether to add related query params to the paramstring
     * action: the action to perform. can be info|play|text or a Kodi builtin
+    * close: override window close and execute a Kodi builtin instead
 window_id=window_id(int)
     * The custom window that will act as the base window
 setproperty=property(str)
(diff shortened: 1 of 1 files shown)


roidy:
Thanks, after testing I can confirm both methods work perfectly 👍

jurialmunkey:
@roidy - Thanks for testing. Did you want me to leave it open for a bit before merging to test it out a bit more in case of anything unintended? Or do you feel confident it is not causing any other issues?

I can't think of anything where it would be a problem, but you'd also be in a better position to test than me since I haven't used the recommendations script in a while. 

I'm not fussed either way (happy to merge now or wait) - just don't want to jump the gun if you want to test it out a bit more first!

roidy:
I'll test it for a couple more day just to make sure, no point merging it and then me finding a issue later.

> I haven't used the recommendations script in a while.

Which method do you use? Is there now a better method?

Thanks

jurialmunkey:
> Which method do you use? Is there now a better method?

I just use a custom dialog with the 1190 window
(i.e. basically the old method but I use the TMDbHelper.DirectCallAuto setting to open info dialog directly).

It's not better or worse, just more useful for my needs (I split the dialogs - one for director/writer/studio, one for year/genre/recs, and one for cast member details), plus I just generally prefer native dialogs vs. scripted ones.

Most of the speed advantage in the recs script comes from the fact that it only does the TMDb ID lookup once for a query vs. multiple times for individual containers in a custom dialog; so I get most of the same benefit by making sure I pass the TMDb ID from service monitor to a property in the custom dialog before I activate the window and then I only use `tmdb_id=` in my dialog lists and never use a query. It's probably not quite as fast as recs script still but gets much closer and lets me split it across different dialogs.

roidy:
I spent most of yesterday working on my recommendations dialog and testing this and didn't run into any issues, I think it can be safely merged 👍

Issue #1080, opened 2023-07-16
[BUG] Metadata and posters dissapering 


SJ1O:
### Describe the bug

When I play a movie via a tmdb helper widget I don't have a poster or any meta information in the OSD Display.

However if I go into tmdb helper and play the movie from there I do get the poster and meta information in the OSD Display.

Is there a setting I need to change or anything I can do to fix this? 

I have tried it on several skins and the issue remains and I have tried it with two players and the issue remains.

Also when in the tmdb add on if I click a movie the artwork and metainfo disappeares if I refresh details the information comes back I'm not sure if this issue is related to just thought I'd mention it.

I have included some photos.

Also after some extensive testing I have noticed the issue only happens when player URL resolver is set to true.

Please help me tmdb helper is my favourite add on I use it for everything.

### Steps To Reproduce

Play movie via widget after afew seconds of the movie playing the meta information and poster disappeares.

I'm using trakt watchlist widget.

This issues doesn't happen when going direct via the tmdb helper addon to play a movie however If I click a movie to view it's information then go back the movie information and metadata disappeares. I think this could be related.

### Screenshots and Additional Info

![20230716_012418](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/assets/121204283/0f6df62d-54d0-4ad9-9953-b952ea686a6a)
![20230716_012603](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/assets/121204283/00e531b2-a200-4fb7-8fd9-f9505f2701ed)

### Checklist

- [X] I have searched the issues and this bug has not been reported.
- [X] I have updated TMDbHelper to the most recent version.
- [X] I have described the bug and included steps to reproduce the issue
- [ ] I have uploaded my debug log.

SJ1O:
Here is another photo of the other issue I described about data dissapering after viewing movie information page.
![20230716_014332](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/assets/121204283/c1816c72-4a61-470f-9b80-02136438a206)

jurialmunkey:
What are you playing this content via? TMDbHelper doesn't provide playable content directly. 
Without knowing what addon you're using it is difficult to guess what the issue might be.

>Also after some extensive testing I have noticed the issue only happens when player URL resolver is set to true.

So why not set it to false then? 
It sounds like the addons you are using are not compatible with this setting, which is why it is optional.

[...]

PR #1262, opened 2024-10-12
Update trakt.py

Peter3344:
its fix the disappearing of metadata when enabling get resume point from trakt this should fix isuue #1080

### DIFF
--- a/resources/tmdbhelper/lib/items/trakt.py
+++ b/resources/tmdbhelper/lib/items/trakt.py
@@ -39,7 +39,7 @@ class TraktMethods():
             progress = 0
         li.infoproperties['ResumeTime'] = int(duration * progress // 100)
         li.infoproperties['TotalTime'] = int(duration)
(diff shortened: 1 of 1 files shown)


jurialmunkey:
This does not fix anything. `set_playprogress` requires a filepath (i.e. the listitem url) not a listitem
https://github.com/jurialmunkey/script.module.jurialmunkey/blob/0dc38f055f3e6c03e5b974d431ca7987611fe7af/resources/modules/jurialmunkey/jsnrpc.py#L110-L113

You're passing the wrong value which effectively breaks the function. The reason it "works" is because, by breaking the function, you're effectively just disabling the setting to pass the resume point back to Kodi.

The bug does not exist in TMDbHelper. The issue is within Kodi about how it manages restore points and the JSON RPC interface for setting them.

Peter3344:
Ok I got you but is it possible to pass also listitem

PR #1294, opened 2024-12-23
Add TV show title to episode labels for improved readability

vkononov:
Previously, episode labels only displayed the episode title, which could be ambiguous—especially for generic titles like "Episode 1" or "Episode 2." Even for more descriptive episode titles, it was difficult to identify the associated TV show. By including the TV show's title in the episode label, this change improves clarity and ensures better context in list views.

See example below that depicts the before and after:

Screenshot 2024-12-17 at 8 49 00 PM

Screenshot 2024-12-17 at 8 49 24 PM

### DIFF
--- a/resources/tmdbhelper/lib/items/listitem.py
+++ b/resources/tmdbhelper/lib/items/listitem.py
@@ -446,11 +446,17 @@ class _Episode(_Tvshow):
             return
         self._set_params_reroute_default()
 
-    def set_episode_label(self, format_label=u'{season}x{episode:0>2}. {label}'):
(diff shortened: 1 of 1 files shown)


jurialmunkey:
I disagree about this change. This would look terrible for episodes in a season that all belong to the same tvshow. It only works for this very specific list but your change is applied globally to all episodes. 

It is up to the skinner if they want to display additional details. This behaviour shouldn't be forced by a plugin. The tvshowtitle is already available to skins as an infolabel if they want to show it in lists.

vkononov:
Fair enough. 😊

PR #1304, opened 2025-01-12
Add JackTook player

Coffelius:
It adds the JackTook player. I'm using it everyday and it works like a charm.

[plugin.video.jacktook](https://github.com/Sam-Max/plugin.video.jacktook)


jurialmunkey:
This sounds like a banned addon. I cannot include players for banned addons.

PR #1331, opened 2025-03-02
Update [DE] strings.po

ksooo:
Rework complete German translation to make it consistent with Kodi German Glossary and translation rules.


jurialmunkey:
lgtm

PR #1342, opened 2025-03-16
Add "Remove rating"

matke-84:
I added to code that when removing a rating, it would say "remove rating" instead of "change rating".

### DIFF
--- a/resources/tmdbhelper/lib/script/sync.py
+++ b/resources/tmdbhelper/lib/script/sync.py
@@ -480,6 +480,13 @@ class _Rating():
                 self._item.season, self._item.episode)
         if not item:
             return
+
+        # We check if there is a rating on the Trakt server
+        had_rating = self._trakt.is_sync(
+            self._item.trakt_type, self._item.unique_id, self._item.season,
+            self._item.episode, self._item.id_type, "ratings"
+        )
+
         # Ask user for rating
         try:
             x = int(Dialog().numeric(0, f'{self.name} (0-10)'))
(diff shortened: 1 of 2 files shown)


resources/tmdbhelper/lib/script/sync.py
```diff
@@ -500,4 +500,12 @@ def sync(self):
                 'ratings/remove' if x == 0 else 'ratings',
                 postdata={f'{self._item.trakt_type}s': [item]}
             )
+
+        if not self._trakt.is_sync(self._item.trakt_type, self._item.unique_id, self._item.season, self._item.episode, self._item.id_type, 'ratings'):
```
jurialmunkey: This feels excessive to do a sync just to check whether to use add or change as the label for the result. It also seems incorrect because the data is posted just above this so you are racing the trakt server to see if the data is there yet.

matke-84: The code works. Somehow you have to check if there is already a rating. You certainly have a lot more experience with your code and with the trakt server and api. Well, if you can do it based on this, it would be nice.

jurialmunkey: But it doesn't work how you think it does. Youre just getting lucky that you're beating the server syncing.

There is a much better easier check against the original name attribute. If name is already the add localisation then it is a new rating, otherwise it is a change rating.

matke-84: Ok. As I told you, do as you think it should be. I cannot know your code better than you. I just wanted to give a some idea to solve this issue that was opened a long time ago.

PR #1366, opened 2025-04-07
Add Chinese Simplified Chinese

wabisabi926:
Add Chinese Simplified Chinese


wabisabi926 (commented):
Update (zh) strings.po

wabisabi926 (commented):
update zh

jurialmunkey:

PR #1367, opened 2025-04-07
Update strings.po

Havokdan:


jurialmunkey:

PR #1373, opened 2025-04-17
Adding Swedish translation

yeager:


jurialmunkey:

You should be still able to use the translated strings for a new translation file -- just the IDs will have some changes (mostly it's a case of several being removed).

PR #1382, opened 2025-04-22
Added Hungarian translation

frodo19:


jurialmunkey:
Unfortunately, need to close this one as the recent refactor means that this file is going to be out of date. A lot of unused strings were cleaned up and removed in the most recent version.

frodo19:
I will try it in next week again, with new pr.

PR #1432, opened 2025-05-06
FR Update strings.po

Bindou01:


resources/language/resource.language.fr_fr/strings.po
```diff
@@ -174,7 +199,7 @@ msgstr "Utiliser des lecteurs groupés"
 #: /resources/settings.xml
 msgctxt "#32038"
 msgid "Get additional details from the local Kodi library database"
-msgstr "Obtenir des informations supplémentaires depuis votre médiathèque locale kodi"
+msgstr "Obtenir des informations supplémentaires depuis la base de données de la médiathèque Kodi locale"
```
Nanomani: If I can say something : 
IMHO the label is a bit too long and with new 3 options this seems to me enought with this : 
"Obtenir des informations depuis votre médiathèque locale kodi"

Then you are 3 options : 
 - Désactivé
 - Préféré
 - Supplément

So it is just my think. Make you want :)
Anyway Thanks for that

Bindou01: C'est effectivement une bonne idée. :)

PR #1469, opened 2025-05-13
Update strings.po

Havokdan:


jurialmunkey:
Looks like this PR conflicts with the other one. I wasn't confident to know which is the correct translation.

PR #1534, opened 2025-05-26
🐛 Resolves missing 'dbid_sorted' attribute error in sync lists

OfficerKD637:
Watchlist, Collection, Favourites, etc. weren't working and giving the missing attribute error when trying to access them inside TMDbH.

This resolves the issue but I'm not sure if it's the correct way. Should it be 'None' or 'False' instead of 'True'?


jurialmunkey:
Ah yeah, it should be False. I think probably should have the default set in the ListProperties class itself (then derived classes can flip to True if they support it).

PR #1619, opened 2025-07-18
🐛 Fix small typo and add back the "Trending Today" list

OfficerKD637:
- Noticed a duplicated string (32179 - "Trending This Week") when accessing the movie directory inside the addon, so matched it with the appropriate list name (32180 - "In Theatres").

- Also added back the "Trending Today" list.


jurialmunkey:
lgtm

PR #1696, opened 2025-11-23
Update strings.po

Havokdan:


resources/language/resource.language.pt_br/strings.po
```diff
@@ -822,12 +826,10 @@ msgstr "Usar fanart.tv para arte do monitor de serviço"
...
 msgctxt "#32164"
-msgid ""
-"Adding Trakt lists to library is limited to [B]{}[/B] TV Shows and [B]{}[/B] Movies per list.\n"
+msgid "Adding Trakt lists to library is limited to [B]{}[/B] TV Shows and [B]{}[/B] Movies per list.\n"
 "This limit is set to ensure that the APIs providing data for free are respected."
-msgstr ""
-"A adição de listas Trakt à biblioteca é limitada a [B]{}[/B] programas de TV e [B]{}[/B] filmes por lista.\n"
-"Esse limite é definido para garantir que as APIs que fornecem dados gratuitamente sejam respeitadas."
+msgstr "A adição de listas do Trakt à biblioteca é limitada a [B]{}[/B] programas de TV e [B]{}[/B] filmes por lista.\n"
+“Esse limite é definido para garantir que as APIs que fornecem dados gratuitamente sejam respeitadas.”"
```
jurialmunkey: Quotation marks issue on this line.

resources/language/resource.language.pt_br/strings.po
```diff
@@ -847,12 +849,12 @@ msgstr "[B]{}[/B] pelo usuário [B]{}[/B]"
 #: /resources/settings.xml
 msgctxt "#32169"
 msgid "Logout of Trakt account..."
-msgstr "Sair da conta do Trakt..."
+msgstr "Sair da conta Trakt...""
```
jurialmunkey: Doubled up quotation marks on end of this line.

jurialmunkey (changes requested):

resources/language/resource.language.pt_br/strings.po
Havokdan: I hope I fixed it, I'm not very familiar with Github functions, but I think I fixed it.

resources/language/resource.language.pt_br/strings.po
Havokdan: I hope I fixed it, I'm not very familiar with Github functions, but I think I fixed it.

resources/language/resource.language.pt_br/strings.po
jurialmunkey: I think you've half fixed it. Error is because you've got the wrong quotation marks character at the start of the line. You've fixed the one at the end but not the one at the start of the line.

You've got one of these `“` when it should be one of these `"`

It's very subtle difference but hopefully you can see what I mean. I'm guessing something in your text editor has changed it from a normal double-quote to a fancy one

PR #1733, opened 2025-12-19
update Chinese

wabisabi926:


resources/language/resource.language.en_gb/strings.po
jurialmunkey, line 1: This file should not be modified.

PR #1751, opened 2026-01-11
add spanish addon.xml

01xKeven:


addon.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon id="plugin.video.themoviedb.helper"
-version="6.14.7"
+version="6.14.8"
```
jurialmunkey: Dont bump version.

jurialmunkey (changes requested):

addon.xml
01xKeven: Done

PR #1758, opened 2026-01-16
Fix TMBD lookups in PVR submenus

Lam4hound:
TMDB lookups were not working in three menus (recordings, search and timers).

Tested in Arctic Fuse 2 & 3.


jurialmunkey (changes requested):
Check for MyPVRTimers.xml also needs to be added here:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/blob/150b3d3b2869fcf901a85ea420ea4eee2939bf96/resources/tmdbhelper/lib/monitor/itemdetails.py#L112-L119

Lam4hound:
I edited the itemdetails.py. Also changed the order of the MyPVR*.xmls to match.

PR #1774, opened 2026-01-26
Add localonly support for Trakt lists (extends localitems branch)

dangerouslaser:
## Summary

This PR has been reworked to adopt the architecture from the `localitems` branch and extend it with Trakt support.

**What changed from the original PR:**
- Dropped the old `library_only` implementation entirely (addressed all review feedback)
- Cherry-picked jurialmunkey's `localitems` commit (`e3d244ec`) as the base
- Added Trakt standard and filtered list support on top using the same patterns

**New code adds:**
- `ListTraktStandardLocalProperties` — generator-based filtering with `kodi_db.get_info()` matching (same pattern as TMDb `ListStandardLocalProperties`)
- Extracted `final_items` cached_property from `ListTraktStandardProperties`
- Refactored `ListTraktFilteredProperties` into a mixin (matching the `ListDiscoverProperties` refactor)
- `ListTraktFilteredStandardProperties` and `ListTraktFilteredLocalProperties` via multiple inheritance
- `@property` factory on `ListTraktStandard.list_properties_class` and `ListTraktFiltered.list_properties_class`

## Review feedback addressed

| Comment | Resolution |
|---------|-----------|
| "Why cast to str?" | `get_tmdb_ids()` removed entirely — matching uses `kodi_db.get_info()` instead |
| "10 pages x 10 widgets = rate limiting" | `page_limit = 8` with early exit via generator |
| "Bad fallback to unfiltered results" | Generator returns empty list when no matches |
| "Should be its own class via factory" | Separate `*LocalProperties` classes selected via `@property` factory |
| "Code duplication / use inheritance" | `ListTraktStandardLocalProperties` inherits from `ListTraktStandardProperties` — same pattern, no duplication |
| "Applying globally for something which only exists for some lists" | Only list types whose `list_properties_class` property checks `is_localonly` support it |

## Usage
```
plugin://plugin.video.themoviedb.helper/?info=popular&tmdb_type=movie&localonly=true
```

Works with TMDb standard lists, TMDb discover lists, Trakt standard lists, and Trakt filtered lists (trending, popular, most played, most watched, anticipated).

## Test plan
- [ ] Test widget with `localonly=true` shows only items in local Kodi library
- [ ] Test widget without param works normally (no behavior change)
- [ ] Test with movie and tv content types
- [ ] Test TMDb standard lists (popular, trending, top_rated, etc.)
- [ ] Test TMDb discover lists
- [ ] Test Trakt standard lists (box office, recommendations)
- [ ] Test Trakt filtered lists (trending, popular, most played, most watched, anticipated)
- [ ] Verify pagination is naturally disabled via `next_page` override
- [ ] Verify cache keys are distinct between standard and local variants

### DIFF
--- a/resources/tmdbhelper/lib/items/container.py
+++ b/resources/tmdbhelper/lib/items/container.py
@@ -60,6 +60,10 @@ class ContainerDirectoryCommon(CommonContainerAPIs):
         return boolean(self.params.get('cacheonly', self.default_cacheonly))
 
     @cached_property
+    def is_localonly(self):
+        return boolean(self.params.get('localonly', False))
+
+    @cached_property
     def is_detailed(self):
         if self.params.get('info') == 'details':
             return True
--- a/resources/tmdbhelper/lib/items/directories/tmdb/lists_standard.py
+++ b/resources/tmdbhelper/lib/items/directories/tmdb/lists_standard.py
@@ -1,5 +1,6 @@
 from tmdbhelper.lib.items.directories.lists_default import UncachedItemsPage, ListDefault, ListProperties
 from tmdbhelper.lib.addon.plugin import get_setting
+from jurialmunkey.ftools import cached_property
 
 
 class ListStandardProperties(ListProperties):
@@ -15,12 +16,16 @@ class ListStandardProperties(ListProperties):
     def get_api_response(self, page=1):
         return self.tmdb_api.get_response_json(self.url, page=page)
 
+    @cached_property
+    def final_items(self):
+        return [
+            i for xpage in range(self.page, self.next_page)
+            for i in self.class_pages(self, xpage).items
+        ]
+
     def get_uncached_items(self):
         return {
-            'items': [
-                i for xpage in range(self.page, self.next_page)
-                for i in self.class_pages(self, xpage).items
-            ],
+            'items': self.final_items,
             'pages': self.total_pages,
             'count': self.total_items,
         }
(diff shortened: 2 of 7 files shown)


dangerouslaser:
## Testing Completed

Tested on Kodi instance (192.168.1.33) with:
- **3,816 movies** with TMDb IDs in local library
- **889 TV shows** with TMDb IDs in local library

### Test Results

| Test | Result |
|------|--------|
| Widget with `library_only=true` (TMDb movies) |  No errors |
| Widget without `library_only` param |  No errors |
| TV shows with `library_only=true` |  No errors |
| Trakt popular with `library_only=true` |  No errors |

### Log Analysis
- No Python tracebacks or exceptions
- No TMDbHelper-related errors
- Plugin windows loaded successfully (MyVideoNav.xml)

All tests executed without errors. The plugin opens and displays content correctly with the `library_only` parameter.

dangerouslaser:
@jurialmunkey I split a new branch from my original fork to break this PR down based on your feedback. This is just the first feature. I tested on a local kodi install and a coreelec install.

Pagination needs to be disabled for this to work as the filtered widgets seem to ignore the global pagination (next page item) setting, and it's very likely that for widgets like "in theaters" there will be less than 20 items matching a users library. if you have any thoughts on this I can revisit.

Let me know if you have any questions.

resources/tmdbhelper/lib/api/kodi/rpc.py
```diff
@@ -182,6 +182,12 @@ def _get_kodi_database():
 
         return database
 
+    def get_tmdb_ids(self):
+        """Return set of all TMDb IDs in the library for fast lookup."""
+        if not self.database:
+            return set()
+        return {str(item['tmdb_id']) for item in self.database if item.get('tmdb_id')}
```
jurialmunkey: Why is this cast to str ?

resources/tmdbhelper/lib/items/directories/tmdb/lists_standard.py
```diff
@@ -1,6 +1,9 @@
 from tmdbhelper.lib.items.directories.lists_default import UncachedItemsPage, ListDefault, ListProperties
 from tmdbhelper.lib.addon.plugin import get_setting
 
+# Constants for library-only mode
+LIBRARY_ONLY_MAX_PAGES = 10
```
jurialmunkey: Could be problematic with lots of widgets

10 widgets x 10 pages is 100 simultaneous requests. Very likely to cause rate limiting.

resources/tmdbhelper/lib/items/directories/tmdb/lists_standard.py
```diff
@@ -16,13 +19,52 @@ def get_api_response(self, page=1):
         return self.tmdb_api.get_response_json(self.url, page=page)
 
     def get_uncached_items(self):
+        # Standard behavior when library_only is not active
+        if not self.is_library_only or not self.library_tmdb_ids:
```
jurialmunkey: Seems a bad idea to fallback to unfiltered results if no matches.

Method is also doing too many things. The library only approach should be its own class and then generated via factory. I don't think you're going to get good clean code vibe coding with Claude.

resources/tmdbhelper/lib/items/directories/tmdb/lists_standard.py
```diff
@@ -16,13 +19,52 @@ def get_api_response(self, page=1):
...
         return {
-            'items': [
-                i for xpage in range(self.page, self.next_page)
-                for i in self.class_pages(self, xpage).items
-            ],
-            'pages': self.total_pages,
-            'count': self.total_items,
+            'items': matching_items,
+            'pages': 0,  # No pagination for library-only
+            'count': len(matching_items),
```
jurialmunkey: Way too much going on here for a single method.

resources/tmdbhelper/lib/items/directories/trakt/lists_standard.py
```diff
@@ -67,10 +70,49 @@ def url(self):
...
+                    if len(matching_items) >= self.library_target_count:
+                        break
+
+            pages_fetched += 1
+            current_page += 1
+
+            # Stop if we have enough items
+            if len(matching_items) >= self.library_target_count:
+                break
+
```
jurialmunkey: A lot of unnecessary duplication here. Need to separate into classes that can be inherited instead of duplicating code. Again, don't think you're going to get good code vibe coding with Claude.

resources/tmdbhelper/lib/items/container.py
```diff
@@ -97,6 +102,8 @@ def pagination(self):
             return False
         if self.is_widget and not get_setting('widgets_nextpage'):
             return False
+        if self.is_library_only:
```
jurialmunkey: What about lists that don't support this param. This is applying globally for something which only exists for some lists.

jurialmunkey (commented):
While this code might "work" and has been tested, it's not very clean. There should be proper consideration of classes and inheritance and single purpose methods.

jurialmunkey:
Here is a proper class based method with methods limited to one concern and using a factory to determine which class to use. Uses `localonly=true` param. Currently works for basic TMDb lists and TMDb discover but would be trivial to adapt to Trakt lists.

https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/e3d244ecf191925cecb5f292e84ecd8b68e9afc1

I'm not certain, however, if this general approach of collecting pages is the best method. It seems like a quite wasteful brute force method to achieve this.

dangerouslaser:
> Here is a proper class based method with methods limited to one concern and using a factory to determine which class to use. Uses `localonly=true` param. Currently works for basic TMDb lists and TMDb discover but would be trivial to adapt to Trakt lists.
> 
> [e3d244e](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/e3d244ecf191925cecb5f292e84ecd8b68e9afc1)
> 
> I'm not certain, however, if this general approach of collecting pages is the best method. It seems like a quite wasteful brute force method to achieve this.

Thank you @jurialmunkey 

The thought process for parsing 10 pages was that for playlists such as In Theaters and On The Air it, it would be unlikely for a user to have 20 of those items in the first page of results on TMDB. I found that even with my expansive library I was only getting 5-10 results for each. I was trying to get to 20 results for a "fuller" widget, but there may be a better way to do this. I am working on an addon that lives separate from TMDB helper which may be a better approach as I am really just trying to use claude to put together an MVP and then refactor.

PR #1824, opened 2026-05-26
Honor vacuum/delete intervals (restore @property semantics)

pannal:
## Summary

`DatabaseMaintenance.is_next_vacuum` and `is_next_delete` are defined as regular methods but consumed as bare attribute lookups (`not self.is_next_vacuum`, no parens). Python evaluates `not <bound_method>` as `False`, so the guards in `vacuum()` and `delete_legacy_folders()` never trip. Both run on every `CronJobMonitor` tick (~10 minutes) instead of the intended 6h / 24h intervals.

Two fixes in one commit:

1. Decorate `is_next_vacuum` and `is_next_delete` with `@property` so the existing call sites actually evaluate the timestamp.
2. Fix `delete_legacy_folders()` checking `is_next_vacuum` instead of `is_next_delete`. Typo, masked until now by the `not <method>` bug.

## Why

Found while tracing a microstutter report (see #1823). The debug log shows the vacuum running every ten minutes for the entire 5-hour session, e.g.:

```
13:40:34  Vacuuming databases: 3.724 sec
13:50:38  Vacuuming databases: 3.735 sec
14:00:41  Vacuuming databases: 3.539 sec
14:10:45  Vacuuming databases: 3.661 sec
14:20:49  Vacuuming databases: 3.692 sec
14:30:56  Vacuuming databases: 7.401 sec
```

With these fixes, vacuum runs at most once per `vacuum_interval` (6h), and legacy-folder cleanup runs at most once per `delete_interval` (24h).

## Test plan

- [x] After a fresh start, vacuum runs once on the first tick then waits `vacuum_interval` before running again.
- [x] `delete_legacy_folders()` only runs once per `delete_interval`, not on every tick.
- [x] `force=True` callers still run unconditionally.

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon id="plugin.video.themoviedb.helper"
-version="6.15.6"
+version="6.15.7"
 name="TMDb Helper"
 provider-name="jurialmunkey">
 <requires>
--- a/resources/tmdbhelper/lib/script/method/maintenance.py
+++ b/resources/tmdbhelper/lib/script/method/maintenance.py
@@ -36,10 +36,12 @@ class DatabaseMaintenance:
         from tmdbhelper.lib.addon.tmdate import set_timestamp
         self.next_delete = set_timestamp(self.delete_interval)
 
+    @property
     def is_next_vacuum(self):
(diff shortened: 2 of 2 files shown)


jurialmunkey:
Ah I just fixed this and then saw this PR after. I could have just merged this one!

PR #1840, opened 2026-07-19
Fix TypeError: NoneType comparison in is_inprogress_show

aravindtri:
is_inprogress_show() crashes with TypeError when Trakt sync returns null for watched_episodes or aired_episodes. Add None to 0 fallback guards.


jurialmunkey:
I dont think this is the correct way to deal with this. None means a different thing to 0. The reason for receiving null/None needs to be addressed.

PR #1841, opened 2026-07-19
Fix SQLite connection lifecycle across threads

primez-x:
TMDb Helper was intermittently crashing Kodi during startup. The logs showed SQLite objects being created in one thread and then used from another; on my Kodi CoreELEC devices, this frequently ended in a SIGSEGV inside _sqlite3.so, causing Kodi to frequently crash and restart ~10-15sec after startup.

TMDb Helper cached a SQLite cursor on an object shared by multiple worker threads. It also closed the cursor without reliably closing its underlying connection, leaving cleanup to Python.

Each thread now gets its own cursor and connection. Both are always closed, failed transactions roll back, and database initialization safely handles interrupted migrations.

This fixes the underlying SQLite ownership problem instead of bypassing it with check_same_thread=False.

### DIFF
--- a/resources/tmdbhelper/lib/api/contains.py
+++ b/resources/tmdbhelper/lib/api/contains.py
@@ -1,4 +1,4 @@
-from jurialmunkey.ftools import cached_property
+from jurialmunkey.ftools import cached_property, threaded_cached_property
 
 
 class CommonContainerAPIs():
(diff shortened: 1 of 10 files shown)


jurialmunkey:
Where is the bug report that indicates this issue?

jurialmunkey:
This is a bad approach to "fixing" the issue.

The reused cursor being used in a thread needs to be identified and stopped, not a bunch of excessively defensive blocks on threading.

PR #1843, opened 2026-07-20
Fix non-dictionary values in Trakt stats

### LINKED ISSUE #1842: :bug: trak_stats.py dictionary issue
### Describe the bug

Hi Jurial, while testing a custom player addon I ran into a reproducible issue where TMDb Helper's playback monitor would terminate during the stop/update path. The exception came from trakt_stats.py assuming every top-level value in the stats response was a dictionary and calling .items() unconditionally. Adding an isinstance(base_v, dict) guard resolved the exception and continuous playback/scrobbling resumed normally. Would you be interested in looking at a one-line defensive patch?  thanks for your efforts

### Steps To Reproduce

the description covers this

### Debug log

the log just shows where trak_stats.py crashes when it hits the dictionary issue.

_No response_

### Checklist

- [x] I have searched the issues and this bug has not been reported.
- [x] I have updated TMDbHelper to the most recent version.
- [x] I have described the bug and included steps to reproduce the issue
- [x] I have uploaded my debug log.

Sanjays2402:
Closes #1842

Trakt stats responses can include scalar top-level metadata alongside stat dictionaries. Ignore those scalar values instead of calling `.items()` on them; the regression test covers a mixed mapping-and-string response.

Tested with `python3 -m unittest -v tests/test_trakt_stats.py`.

### DIFF
--- a/resources/tmdbhelper/lib/query/database/trakt_stats.py
+++ b/resources/tmdbhelper/lib/query/database/trakt_stats.py
@@ -78,6 +78,7 @@ class GetTraktStatsRequest:
                 'stat': item_v,
             }
             for base_k, base_v in self.response_json.items()
+            if isinstance(base_v, dict)
             for item_k, item_v in base_v.items()
             if isinstance(item_v, int)
         ] if self.response_json else []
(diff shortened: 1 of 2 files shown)


jurialmunkey:
The bug needs to be verified before fixes are implemented. The Trakt API indicates only dictionaries are returned in this data.

PR #1844, opened 2026-07-21
Support availability filter in trakt discovery

Noustaa:


Noustaa:
I used same labels (strings.po) as the ones on the website:
image

jurialmunkey:
Do you know where this is documented? I couldn't see it in the API docs.

Code looks good from a quick glance, just want to check the docs first before I merge!

Noustaa:
You can see details here (they all have the same watchnow query parameter): 

https://docs.trakt.tv/reference/getmoviestrending
https://docs.trakt.tv/reference/getmoviespopular
https://docs.trakt.tv/reference/getshowstrending
https://docs.trakt.tv/reference/getshowspopular

I didn't implemeneted all the possible values because i do not have them available on the website so i wouldn't be able to actually test them.

jurialmunkey:
Ah I didn't realise they had a new docs website. I was still looking at the old apiary one.

Noustaa:
Do you need anything more to merge it ?

jurialmunkey:
> Do you need anything more to merge it ?

Sorry just been a bit busy. Will review when I get a chance, but looks like it should be OK

PR #1851, opened 2026-07-25
✨ Allow option to input user Trakt Client ID and Secret Key

OfficerKD637:
Need to re-authenticate for the new credentials to take effect.

Can move the setting to the API section if required, but the dedicated Trakt category made more sense.

---

Hey J!

I know you already said that you weren't interested in working on any new Trakt features (or any band-aid workarounds) but I felt this was easy enough to add so decided to make a PR anyway haha.

I haven't had the time to migrate my personal setup over to MDBList, so I'm still hanging on by a thread lol 🤣 

Thank you for your time!

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -2892,6 +2892,26 @@ msgid "You must authenticate your Trakt account to use this feature."
 msgstr ""
 
 #: /resources/settings.xml
+msgctxt "#32403"
+msgid "User Client ID"
+msgstr ""
+
+#: /resources/settings.xml
+msgctxt "#32408"
+msgid "User Client Secret Key"
+msgstr ""
+
+#: /resources/settings.xml
+msgctxt "#32417"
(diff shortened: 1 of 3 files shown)


jurialmunkey:
I don't think this is a good idea to try to workaround their restrictions. Likely would lead to a complete block or them taking more drastic measures.

jurialmunkey:
This would almost certainly break their terms of service.

OfficerKD637:
Oh I see... There is nothing in their ToS (_currently_) that states this is prohibited. I think a lot of open source Trakt apps ([like this one](https://github.com/angristan/trakt-cli)) encourage users to supply their credentials, but you never know with them haha. One bad day away from another shipwreck lol.
I guess the distinction lies in the intention. Having the option to enter your credentials for "expert users" or "development" purposes v/s to get around their restrictions.

jurialmunkey:
"don't create additional API apps to avoid rate limits or other API restrictions we put in place."
https://docs.trakt.tv/docs/create-an-app

OfficerKD637:
Yikes sorry...

PR #1854, opened 2026-07-27
Fix: OSD poster display

LXT-A-X:
The issue is that the OSD call passes extra 
"season" and 
"episode" parameters, causing the plugin to return an 
"episode" type (single episode), whose artwork only includes a landscape still.

However, when browsing externally without passing these two parameters, it returns a 
"tvshow" type with the correct vertical poster.

Modification Plan

Only one core file, 
"lists_details.py", was modified:

In 
"ListDetails.get_items()" and 
"ItemDetails.__init__()", when 
"tmdb_type='tv'" and both 
"season" and 
"episode" parameters are passed simultaneously, these two parameters are directly ignored (set to 
"None"). This way:

* 
"self.lidc.get_item('tv', tmdb_id, None, None)" retrieves the details of the entire TV show (tvshow), exactly the same as clicking details externally.
* The returned 
"mediatype='tvshow'" (not 
"episode"), and the 
"art" contains the correct vertical 
"poster".
* Kodi opens the TV show information dialog, and the poster area normally displays the vertical poster.
* 
"container_content" is also 
"tvshows", consistent with the external details page.

At the same time, the defensive modification in 
"listitem.py"'s 
"_Episode" class (falling back 
"poster" to 
"tvshow.poster") has been retained as a safety net for other scenarios.

Screenshot_2026-07-26-23-02-31-306_com tencent mobileqq
Image_1785117183648_774
IMG_20260726_222249

### DIFF
--- a/resources/tmdbhelper/lib/items/directories/base/lists_details.py
+++ b/resources/tmdbhelper/lib/items/directories/base/lists_details.py
@@ -5,11 +5,17 @@ from jurialmunkey.ftools import cached_property
 
 class ItemDetails:
     def __init__(self, directory, tmdb_type, tmdb_id, season=None, episode=None):
-        self.directory = directory  # Directory container class instance
+        self.directory = directory
         self.tmdb_type = tmdb_type
         self.tmdb_id = tmdb_id
-        self.season = season
-        self.episode = episode
+        self._season = season
+        self._episode = episode
+        if tmdb_type == 'tv' and season is not None and episode is not None:
+            self.season = None
+            self.episode = None
+        else:
+            self.season = season
+            self.episode = episode
 
     @property
     def lidc(self):
@@ -26,12 +32,10 @@ class ItemDetails:
             'tmdb_type': self.tmdb_type,
             'tmdb_id': self.tmdb_id
         }
-        if self.season is None:
-            return params
-        params['season'] = self.season
-        if self.episode is None:
-            return params
-        params['episode'] = self.episode
+        if self.season is not None:
+            params['season'] = self.season
+        if self.episode is not None:
+            params['episode'] = self.episode
         return params
 
     @cached_property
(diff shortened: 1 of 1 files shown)


OfficerKD637:
Hey there!

Apologies in advance if I'm misunderstanding, but I am a bit confused as to what this is trying to achieve.

Wouldn't removing the season & episode values lead to showing the TV Show info and not the info for the specific episode you're watching? (the different plot would be the biggest give away).

I can see you're using AF3 from the screenshots. This is the [variable](https://github.com/jurialmunkey/skin.arctic.fuse.3/blob/f03aceaee614bd7108b648711362765c0d60aa36/1080i/script-skinvariables-images-includes.xml#L162-L169) for that poster in the info dialog 
```xml
<variable name="Image_Poster">
    <value condition="!String.IsEmpty(ListItem.Property(override_poster))">$INFO[ListItem.Property(override_poster)]</value>
    <value condition="!String.IsEmpty(ListItem.Art(poster))">$INFO[ListItem.Art(poster)]</value>
    <value condition="!String.IsEmpty(ListItem.Art(season.poster))">$INFO[ListItem.Art(season.poster)]</value>
    <value condition="!String.IsEmpty(ListItem.Art(tvshow.poster))">$INFO[ListItem.Art(tvshow.poster)]</value>
    <value condition="!String.IsEmpty(ListItem.Icon)">$INFO[ListItem.Icon]</value>
    <value condition="!String.IsEmpty(ListItem.Property(fallback.icon))">$INFO[ListItem.Property(fallback.icon)]</value>
</variable>
```

As you'll see it has checks for `ListItem.Art(poster)`, `Art(season.poster)`, `Art(tvshow.poster)`, etc. which means it should automatically display the season or TV show poster if available.

Here's what it looks like on my setup:
image

Is this a local file or are you using an addon? If it's a video addon, perhaps it's not passing all the TMDb info to the player correctly?

LXT-A-X:
Indeed. Should reference episode data, and the poster wall should use season images.
Screenshot_2026-07-27-17-49-42-525_com miui gallery

LXT-A-X:
I was able to display it before. After the version was updated, I couldn't see any posters on the osd. There were only LOGO background pictures and art pictures, so it was always empty. I don't know why other places can display movies correctly, but TV dramas can't.

PR #1864, opened 2026-08-12
Add MDBList next episodes and watchlist views

eengert:
## What changed

- add MDBList `Next Episodes` under TV Shows using the `/upnext` endpoint
- add MDBList movie Watchlist, Watchlist (Released), and Watchlist (Anticipated) entries under Movies
- support MDBList's full list sorting menu for the new watchlist routes
- filter released and anticipated watchlists server-side with `released_to` / `released_from`
- map MDBList's nested Up Next response into TMDBHelper episode items for normal TMDb enrichment
- fix MDBList offset pagination to account for the configured multi-page request size

## Why

MDBList currently appears in TMDBHelper primarily as a user-list provider. These additions make its watchlist and next-episode APIs usable as direct widget sources, paralleling the most important Trakt routes after Trakt's API restrictions.

## Validation

- `python3 -m compileall` on all changed modules
- `git diff --check`
- live authenticated smoke test against MDBList `/upnext`, including conversion to a Kodi episode item
- verified watchlist filtering and sorting parameters against the current MDBList OpenAPI schema

### DIFF
--- a/resources/tmdbhelper/lib/addon/consts.py
+++ b/resources/tmdbhelper/lib/addon/consts.py
@@ -68,7 +68,15 @@ PLAYERS_BASEDIR_TEMPLATES = 'special://home/addons/plugin.video.themoviedb.helpe
 PLAYERS_PRIORITY = 1000
 PLAYERS_CHOSEN_DEFAULTS_FILENAME = 'player_defaults'
 
-NO_UNAIRED_LABEL = ('details', 'trakt_calendar', 'library_nextaired', 'videos', 'trakt_watchlist_anticipated', 'trakt_anticipated')
+NO_UNAIRED_LABEL = (
+    'details',
+    'trakt_calendar',
+    'library_nextaired',
+    'videos',
+    'trakt_watchlist_anticipated',
+    'mdblist_watchlist_anticipated',
+    'trakt_anticipated',
+)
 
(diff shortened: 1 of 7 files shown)


jurialmunkey:
Thanks but the plan here is to integrate MDbList data directly into the sync data so that there is no such thing as an "MDbList Watchlist" or a "Trakt Watchlist" etc. -- instead the source of the data to sync with is chosen and the watchlist functions regardless of source. You can already see the beginnings of this implementation in the "List Services" submenu in the most recent git master where you can select watchlist and collection data sync. This will be extended to other data points to allow for watched history and progress.

Implementing direct calls to the API for generating a list is not worthwhile if the data is already synced anyway in the database - you're just making unnnecessary extra API calls when the data already exists locally.

eengert:
Yeah, that makes more sense.  That's a better large-scale approach.

matthius7:
> Thanks but the plan here is to integrate MDbList data directly into the sync data so that there is no such thing as an "MDbList Watchlist" or a "Trakt Watchlist" etc. -- instead the source of the data to sync with is chosen and the watchlist functions regardless of source. You can already see the beginnings of this implementation in the "List Services" submenu in the most recent git master where you can select watchlist and collection data sync. This will be extended to other data points to allow for watched history and progress.
> 
> Implementing direct calls to the API for generating a list is not worthwhile if the data is already synced anyway in the database - you're just making unnnecessary extra API calls when the data already exists locally.

Where do we authorize mdb list ?

PR #1892, opened 2026-09-04
Refresh MDBList activity cache and widgets after playback

eengert:
When MDBList has recorded an episode as watched, TMDb Helper's Next Episodes widget can continue displaying it until a later activity check or manual sync.

This clears `LASTACTIVITIES_DATA` and updates `Widgets.Reload` from `PlayerMonitor.onPlayBackEnded()` and `onPlayBackStopped()`, after stopping the scrobbler. It applies when MDBList is selected as the watched source and MDBList scrobbling is enabled. Using the playback callbacks also covers external players whose playback does not reach TMDb Helper's scrobbler update hook.

The next widget request checks current activities through the existing sync logic. Database checkpoints are preserved so incremental requests can continue using `since`; the callback does not invoke `SyncInvalidator` or force a full watched-history resync.

Validation: Python compilation and `git diff --check` pass. This matches the locally tested 6.16.907 build: with Red Light and an Arctic Fuse 3 TMDb Helper Next Episodes widget, an episode already recorded as watched in MDBList disappeared after playback stopped. This change refreshes local data; it does not address a player failing to submit a watched update to MDBList.

### DIFF
--- a/resources/tmdbhelper/lib/monitor/player.py
+++ b/resources/tmdbhelper/lib/monitor/player.py
@@ -208,11 +208,25 @@ class PlayerMonitor(Player, CommonMonitorFunctions):
 
     def onPlayBackEnded(self):
         self.scrobbler_stop()
+        self.refresh_mdblist_watched()
         self.reset_properties()
 
     def onPlayBackStopped(self):
         self.scrobbler_stop()
+        self.refresh_mdblist_watched()
         self.reset_properties()
+
+    def refresh_mdblist_watched(self):
+        if get_setting('sync_source_watched', 'str') != 'MDbList':
+            return
+        if not get_setting('mdblist_scrobbling'):
+            return
+        from tmdbhelper.lib.addon.consts import LASTACTIVITIES_DATA
+        from tmdbhelper.lib.addon.logger import kodi_log
+        from tmdbhelper.lib.addon.tmdate import set_timestamp
+        self.get_property(LASTACTIVITIES_DATA, clear_property=True)
+        self.get_property('Widgets.Reload', set_property=f'{set_timestamp(0, True)}')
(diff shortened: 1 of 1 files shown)


jurialmunkey:
Probably not a good idea as this will involve a full resync which is much slower for MDbList. 

Depending on how much watched history you have might involve several API calls everytime you hit this (I think my watched history takes about 16 API calls just for episodes).

eengert:
Ok, I'm just trying to fix the new issue I'm seeing with MDBList as the scrobbler - When the most recent available episodes finishes playing, the episode doesn't get removed from the Next Episodes widget (it seems to be removed at some later point, maybe after watching an episode of a different show or after manually syncing the database).

I just tested a different approach locally that avoids running SyncInvalidator('watchedprogress').
I also found that external players like Red Light don’t initialize TMDb Helper’s PlayerScrobbler, so clearing the activity data from there never actually ran.
What worked was clearing only LASTACTIVITIES_DATA when playback ends or stops, then updating Widgets.Reload. That makes the widget check sync/last_activities again, but it only syncs the watched route if MDBList’s episode_watched_at timestamp has actually changed. So it doesn’t force a full watched-history sync every time.
I tested it with Red Light and my MDBList Next Episodes widget, and the completed episode was removed when I returned to the home screen.
I haven’t updated the PR branch yet. Does this approach sound better?

jurialmunkey:
Yeah clearing the lastactivities property would be better (and what I was going to suggest) because then there is still a cached value which can be used for the since property and only pull in newest.

The main issue is the MDbList forces api calls to be consecutive via next cursor so it can be a bit slow to do a full update versus with Trakt where the extra pages can be retrieved using an offset in parallel.

jurialmunkey:
Ah damn I didn't see that you made the update - I could have included it in the update to the repo I just pushed!

Changes lgtm.

PR #1902, opened 2026-09-14
⚡ Check airedepisodes for Seasons with no date

### LINKED ISSUE #1899: :bug: One Piece (1999) not populating all seasons
### Describe the bug

When browsing seasons it shows 5 seasons instead of 23.

### Steps To Reproduce

Search One Piece (1999) via TMDB Helper and browse seasons.

### Debug log

[kodi.log](https://github.com/user-attachments/files/32101841/kodi.log)

Image

### Checklist

- [x] I have searched the issues and this bug has not been reported.
- [x] I have updated TMDbHelper to the most recent version.
- [x] I have described the bug and included steps to reproduce the issue
- [x] I have uploaded my debug log.
- [x] I have read the note on AI usage in reports in https://github.com/jurialmunkey/plugin.video.themoviedb.helper/issues/1856

OfficerKD637:
Resolves #1899

Hey Jurial,

I _hope_ this is the correct approach haha. More info in this [comment](https://github.com/jurialmunkey/plugin.video.themoviedb.helper/issues/1899#issuecomment-5661903170)

Thanks!


jurialmunkey:
changes for seasons should be done on `_Season` class not rerouted in the parent class using a mediatype check.

OfficerKD637:
Sorry about the stupid mistake (don't know how I missed the different subclasses). Is this okay now?

resources/tmdbhelper/lib/items/listitem.py
```diff
@@ -558,6 +558,12 @@ class _Season(_Tvshow):
     mediatype = 'season'
     trakt_type = 'season'
 
+    @cached_property
+    def is_unaired(self):
+        if not self.infolabels.get('premiered'):
+            return not self.infoproperties.get('airedepisodes')
```
jurialmunkey: you can use the `self.airedepisodes` property here -- I know it effectively does the same thing but will be a bit more robust for future changes (e.g. if wanted to add extra logic to calculating airedepisodes then it is all changed from one place)

jurialmunkey (changes requested):

resources/tmdbhelper/lib/items/listitem.py
OfficerKD637: Makes sense. Fixed in https://github.com/jurialmunkey/plugin.video.themoviedb.helper/commit/344b665980ca1cbe20853d775746b2c486923b8a
Thanks!

PR #1905, opened 2026-09-19
Update (PL) strings.po

3vtv:


3vtv:
This replaces my previous PR, now with a single commit.

